PR #1, opened 2019-03-14
Add endpoints to sync 

vkumra-broad:


welder.py
```diff
@@ -0,0 +1,50 @@
+import subprocess, os
+from flask import (Flask, request, abort, jsonify)
+
+app = Flask(__name__)  # create the application instance
+
+
+@app.route('/welder/api/syncTo', methods=['POST'])
+def sync_to():
+    if not request or 'destination' not in request.json:
+        abort(400)
```
helgridly: do you want a return statement after this? right now i think you'd return a 400 to the user and then continue executing this function

welder.py
```diff
@@ -0,0 +1,50 @@
...
+    destination = request.json['destination']
+    source = os.getenv('WELDERVOLUME','')
+    if not validate_bucket(destination):
+        return jsonify({"Destination is not a valid bucket":destination}), 400
+    if (source == ""):
+        return "Source not defined", 400
+    command = ['gsutil', '-m', 'rsync', '-r', source, destination]
+    output = ""
+    for line in run_command(command):
+        output += line
```
helgridly: try `output = "separator".join(array)` instead of a loop here

helgridly: though what you're doing is building a giant string in memory... maybe you want to write the output to a file, and then find a way to stream that file into the response.

welder.py
```diff
@@ -0,0 +1,50 @@
...
+    output = ""
+    for line in run_command(command):
+        output += line 
+    return jsonify({'output':output}), 200
+
+
+@app.route('/welder/api/syncFrom', methods=['POST'])
+def sync_from():
+    if not request or 'source' not in request.json:
+        abort(400)
```
helgridly: same again

welder.py
```diff
@@ -0,0 +1,50 @@
...
+    source = request.json['source']
+    destination = os.getenv('WELDERVOLUME','')
+    if not validate_bucket(source):
```
helgridly: duplicated code here?

welder.py
```diff
@@ -0,0 +1,50 @@
...
+        return "Destination not defined", 400
+    command = ['gsutil', '-m', 'rsync', '-r', source, destination]
+    output = ""
+    for line in run_command(command):
+        output += line 
+    return jsonify({'output':output}), 200
+
+
+def validate_bucket(bucket_name):
+    return bucket_name.startswith('gs:')
```
helgridly: `gs://`

welder.py
vkumra-broad: I'll add the return for completeness but this is working as intended 🤷‍♂️

helgridly: i think i wanna see documentation saying that flask throws an exception and stops executing of this function in order for me to be convinced here

welder.py
```diff
@@ -0,0 +1,52 @@
...
+        return 'Destination not defined', 400
+    return run_command(source, destination)
+
+
+def validate_bucket(bucket_name):
+    return bucket_name.startswith('gs://')
+
+
+def run_command(source, destination):
+    command = ['gsutil', '-m', 'rsync', '-r', source, destination]
```
MatthewBemis: a few things that we should be doing:

we should be running rsync with the `-P` parameter so that it preserves the permissions attributes. localized files will need this in order to be rewritten by RStudio.

in that same boat, because directories are not gcs objects but will still be touched by RStudio, we need to run something like `chown -R 1000:1000 .` to assign the correct ownership to all of the directories when we are doing localization.

i wonder if the welder container will need to be run with elevated privileges in order to run `chown`...

welder.py
```diff
@@ -0,0 +1,52 @@
...
+    output = ''
+    p = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+    output= list((iter(p.stdout.readline, b'')))
+    if('exception' in output[-2].decode('utf-8').lower()):
+        return jsonify({'status':'Error', 'message':output[-2].decode('utf-8').strip()}), 403
+    else:
+        if('Starting synchronization...' in output[-1].decode('utf-8').lower()):
+            return jsonify({'status':'Success', 'message':'Files are up to date.'}), 200
```
MatthewBemis: having `status: Success` in the response body seems redundant. i would prefer that the response codes are the sole indicator of the request succeeding. 

we should also think about the shape of this response and if it's right. 

Maybe a 200 OK response looks something like:

```
{
  "succeeded": 271,
  "failed": 0,
  "totalFiles": 271
}
```

welder.py
```diff
@@ -0,0 +1,52 @@
...
+    return bucket_name.startswith('gs://')
+
+
+def run_command(source, destination):
+    command = ['gsutil', '-m', 'rsync', '-r', source, destination]
+    output = ''
+    p = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+    output= list((iter(p.stdout.readline, b'')))
+    if('exception' in output[-2].decode('utf-8').lower()):
+        return jsonify({'status':'Error', 'message':output[-2].decode('utf-8').strip()}), 403
```
MatthewBemis: is 403 the right error code here? are there other reasons that rsync can fail?

welder.py
helgridly: fwiw i suggested the `status` key, though my opinion isn't super strong. i liked the idea of clients being able to use the same deserializer for both kinds of responses. i think cromwell does a similar thing, though i can't really speak to their reasoning. i abstain my vote here

Qi77Qi:
closing this PR since the repo is repurposed

PR #2, opened 2019-05-10
initial project setup

Qi77Qi:


core/src/main/scala/org/broadinstitute/dsp/workbench/welder/model.scala
```diff
@@ -0,0 +1,12 @@
+package org.broadinstitute.dsp.workbench.welder
+
+import io.circe.Decoder
+
+final case class LocalObjectPath(asString: String) extends AnyVal
+
+object JsonCodec {
+  implicit val localObjectPathDecoder: Decoder[LocalObjectPath] = Decoder.decodeString.emap {
+    str =>
+      if(str.endsWith("ipynb")) Right(LocalObjectPath(str)) else Left("invalid object local path")
```
Qi77Qi: @MatthewBemis Here I'm saying if the local object path doesn't end with `ipynb`, we're failing the request. Is this correct assumption?

MatthewBemis: that assumption isn't correct. it should be possible to create storage links and localize/delocalize files of any type. it's true that for all of our cases up front we'll only be dealing with .ipynb files, but it should be able to support any file types

PR #3, opened 2019-05-15
add /checkMeta, /localize , /safeDelocalize

Qi77Qi:


core/src/main/scala/org/broadinstitute/dsp/workbench/welder/Crc32c.scala
```diff
@@ -0,0 +1,117 @@
+package org.broadinstitute.dsp.workbench.welder
+
+import java.util.zip.Checksum
+
+import cats.effect.{ContextShift, IO}
+import com.google.common.io.BaseEncoding
+import org.broadinstitute.dsde.workbench.google2.Crc32
+import scala.concurrent.ExecutionContext.global
+// Adapted from https://github.com/googlearchive/crc32c-java/blob/master/src/com/google/cloud/Crc32c.java
```
calbach: Is there a reason we can't use this directly? I'd be pretty surprised if there weren't an existing library to produce a crc32 in Scala/Java.

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/model.scala
```diff
@@ -1,9 +1,56 @@
...
+  final case object Live extends SyncStatus {
+    override def toString: String = "LIVE"
+  }
+  // crc32 mismatch
+  final case object Desynchronized extends SyncStatus {
+    override def toString: String = "DESYNCHRONIZED"
+  }
+  // deleted in gcs. (object exists in storagelinks config file but not in in gcs)
+  final case object DeletedRemote extends SyncStatus {
+    override def toString: String = "DELETED_REMOTE"
```
calbach: nit: proposed renaming in my doc to REMOTE_NOT_FOUND

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/model.scala
```diff
@@ -1,9 +1,56 @@
...
+}
+final case class BucketNameAndObjectName(bucketName: GcsBucketName, blobName: GcsBlobName) {
+  override def toString: String = s"${bucketName.value}/${blobName.value}"
+}
 
 object JsonCodec {
   implicit val localObjectPathDecoder: Decoder[LocalObjectPath] = Decoder.decodeString.map(LocalObjectPath)
-}
\ No newline at end of file
+  implicit val localObjectPathEncoder: Encoder[LocalObjectPath] = Encoder.encodeString.contramap(_.asString)
```
calbach: Was swagger codegen not an option? It handles all of these encoding details for you.

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/model.scala
```diff
@@ -1,9 +1,56 @@
...
+  implicit val workbenchEmailEncoder: Encoder[WorkbenchEmail] = Encoder.encodeString.contramap(_.value)
+  implicit val uriEncoder: Encoder[Uri] = Encoder.encodeString.contramap(_.renderString)
+  implicit val uriDecoder: Decoder[Uri] = Decoder.decodeString.emap(s => Uri.fromString(s).leftMap(_.getMessage()))
+  implicit val instanceEncoder: Encoder[Instant] = Encoder.encodeLong.contramap(_.toEpochMilli) //TODO: shall we make this easier for user to read
+  implicit val syncStatusEncoder: Encoder[SyncStatus] = Encoder.encodeString.contramap(_.toString)
+  implicit val gcsBucketNameEncoder: Decoder[GcsBucketName] = Decoder.decodeString.map(GcsBucketName)
+  implicit val gcsBlobNameEncoder: Decoder[GcsBlobName] = Decoder.decodeString.map(GcsBlobName)
+  def parseGsDirectory(str: String): Either[String, BucketNameAndObjectName] = for {
```
calbach: nit: does this belong under JsonCodec? Feels like more of a business logic concern

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/model.scala
```diff
@@ -1,9 +1,56 @@
...
+    parsed <- Either.catchNonFatal(str.split("/")).leftMap(_.getMessage)
+    bucketName <- Either.catchNonFatal(parsed(2))
+      .leftMap(_.getMessage)
+      .ensure("bucketName can't be empty")(s => s.nonEmpty)
+    objectName <- Either.catchNonFatal(parsed.drop(3).mkString("/"))
```
calbach: I think this handles it properly but may be worth a test case:

Input: "gs://bkt/a/b.txt"
Want: ("bkt", "a/b.txt")

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -30,21 +43,80 @@ class ObjectService(googleStorageService: GoogleStorageService[IO]) extends Http
...
+
+    res.compile.drain
+  }
 
-  def checkMeta(req: GetMetadataRequest): IO[Unit] = ???
+  def checkMetadata(req: GetMetadataRequest): IO[MetadataResponse] =
+    for {
+      traceId <- IO(TraceId(randomUUID()))
+      storageLinks <- readJsonFileToA[IO, List[StorageLink]](pathToStorageLinks).compile.lastOrError
+      storageLink = storageLinks.find(_.localBaseDirectory == req.localObjectPath)
```
calbach: path + matching logic needs to be a bit more nuanced, i.e. to handle the `recursive` flag. May be fine for now but should be tracked somewhere.

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -30,21 +43,80 @@ class ObjectService(googleStorageService: GoogleStorageService[IO]) extends Http
...
+      res <- storageLink.fold[IO[MetadataResponse]](IO.raiseError(NotFoundException(s"No storage link found for ${req.localObjectPath}"))) {
+        sl =>
+          for {
+            bucketAndObject <- IO.fromEither(JsonCodec.parseGsDirectory(sl.cloudStorageDirectory.renderString).leftMap(e => InternalException(e)))
+            metadata <- googleStorageService.getObjectMetadata(bucketAndObject.bucketName, bucketAndObject.blobName, Some(traceId)).compile.lastOrError
+            result <- metadata match {
+                        case google2.GetMetadataResponse.NotFound => IO.raiseError(NotFoundException(s"Object(${bucketAndObject}) not found in Google storage"))
```
calbach: This should set a status

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -30,21 +43,80 @@ class ObjectService(googleStorageService: GoogleStorageService[IO]) extends Http
...
+                        case google2.GetMetadataResponse.Metadata(crc32, userDefinedMetadata) =>
+                          for {
+                            isLinkedString <- IO.fromEither(userDefinedMetadata.get("isLinked").toRight(NotFoundException("isLinked key missing from object metadata ")))
```
calbach: isLinked is just supposed to mean whether a there's a matching directory, i.e. set this to false at line 61 and return, rather than throwing an erro

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -30,21 +43,80 @@ class ObjectService(googleStorageService: GoogleStorageService[IO]) extends Http
...
+      entry =>
+        googleStorageService.getObject(entry.bucketNameAndObjectName.bucketName, entry.bucketNameAndObjectName.blobName, None) //get file from google
+          .through(io.file.writeAll(Paths.get(entry.localObjectPath.asString), blockingEc)) //write file to local disk
+    }.parJoin(10)
+
+    res.compile.drain
+  }
 
-  def checkMeta(req: GetMetadataRequest): IO[Unit] = ???
+  def checkMetadata(req: GetMetadataRequest): IO[MetadataResponse] =
```
calbach: execution mode should be somewhere in here, or a TODO

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -30,21 +43,80 @@ class ObjectService(googleStorageService: GoogleStorageService[IO]) extends Http
...
+            metadata <- googleStorageService.getObjectMetadata(bucketAndObject.bucketName, bucketAndObject.blobName, Some(traceId)).compile.lastOrError
+            result <- metadata match {
+                        case google2.GetMetadataResponse.NotFound => IO.raiseError(NotFoundException(s"Object(${bucketAndObject}) not found in Google storage"))
+                        case google2.GetMetadataResponse.Metadata(crc32, userDefinedMetadata) =>
+                          for {
+                            isLinkedString <- IO.fromEither(userDefinedMetadata.get("isLinked").toRight(NotFoundException("isLinked key missing from object metadata ")))
+                            isLinked <- IO(isLinkedString.toBoolean)
+                            fileBody <- googleStorageService.getObject(bucketAndObject.bucketName, bucketAndObject.blobName, Some(traceId)).compile.to[Array]
+                            calculatedCrc32 = Crc32c.calculateCrc32c(fileBody)
+                            syncStatus = if (fileBody.length == 0) SyncStatus.DeletedRemote
```
calbach: I don't think this is how it will work, I would test this out

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -30,21 +43,80 @@ class ObjectService(googleStorageService: GoogleStorageService[IO]) extends Http
...
+  def safeDelocalize(req: SafeDelocalize): IO[Unit] = {
+    for {
+      traceId <- IO(TraceId(randomUUID()))
+      storageLinks <- readJsonFileToA[IO, List[StorageLink]](pathToStorageLinks).compile.lastOrError
+      storageLink = storageLinks.find(_.localBaseDirectory == req.localObjectPath)
+      metadata <- checkMetadata(GetMetadataRequest(req.localObjectPath))
+      _ <- if(metadata.syncStatus == SyncStatus.Live) {
+        io.file.readAll[IO](Paths.get(req.localObjectPath.asString), blockingEc, 4096).compile.to[Array].flatMap {
+          body =>
+            googleStorageService.storeObject(GcsBucketName(""), GcsBlobName(""), body, "text/plain", Some(traceId))
```
calbach: I'm curious - where happens to this traceId?

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -30,21 +43,80 @@ class ObjectService(googleStorageService: GoogleStorageService[IO]) extends Http
...
+          } yield result
+      }
+    } yield res
 
-  def safeDelocalize(req: SafeDelocalize): IO[Unit] = ???
+  def safeDelocalize(req: SafeDelocalize): IO[Unit] = {
+    for {
+      traceId <- IO(TraceId(randomUUID()))
+      storageLinks <- readJsonFileToA[IO, List[StorageLink]](pathToStorageLinks).compile.lastOrError
+      storageLink = storageLinks.find(_.localBaseDirectory == req.localObjectPath)
```
calbach: See recursive comment above

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -30,21 +43,80 @@ class ObjectService(googleStorageService: GoogleStorageService[IO]) extends Http
...
+    }.parJoin(10)
+
+    res.compile.drain
+  }
 
-  def checkMeta(req: GetMetadataRequest): IO[Unit] = ???
+  def checkMetadata(req: GetMetadataRequest): IO[MetadataResponse] =
+    for {
+      traceId <- IO(TraceId(randomUUID()))
+      storageLinks <- readJsonFileToA[IO, List[StorageLink]](pathToStorageLinks).compile.lastOrError
```
calbach: I'm guessing we'd eventually want an interface over the storage links config, and we'd probably want to hold it in memory (with some locking). Not necessarily as part of this PR

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -30,21 +43,80 @@ class ObjectService(googleStorageService: GoogleStorageService[IO]) extends Http
...
+      storageLink = storageLinks.find(_.localBaseDirectory == req.localObjectPath)
+      metadata <- checkMetadata(GetMetadataRequest(req.localObjectPath))
+      _ <- if(metadata.syncStatus == SyncStatus.Live) {
+        io.file.readAll[IO](Paths.get(req.localObjectPath.asString), blockingEc, 4096).compile.to[Array].flatMap {
+          body =>
+            googleStorageService.storeObject(GcsBucketName(""), GcsBlobName(""), body, "text/plain", Some(traceId))
+        }
+      } else {
```
calbach: nit: this should probably throw an error

server/src/main/resources/api-docs.yaml
```diff
@@ -0,0 +1,107 @@
...
+      storageLinks:
+        type: string
+        description: #TODO: Fix this
+  SyncStatus:
+    type: string
+    enum:
+      - LIVE
+      - DESYNCHRONIZED
+      - DELETED_REMOTE
+  ErrorReport:
```
calbach: I would recommend defining this as an object type to allow easier compatibility going forwards (with the current model - adding a new field means breaking old clients).

server/src/main/resources/api-docs.yaml
```diff
@@ -0,0 +1,107 @@
...
+paths:
+  '/objects/metadata':
+    get:
+      summary: 
+      responses:
+        '200':
+          schema:
+            $ref: '#/definitions/Metadata'
+        '404':
+          description: Either requested notebook doesn't exist in google cloud storage, or notebook doesn't have requried metadata associated with it
```
calbach: In the first case, it should still return 200 with "REMOTE_NOT_FOUND" status

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/Crc32c.scala
Qi77Qi: This is the only java library I found(referenced [this page](https://cloud.google.com/storage/docs/hashes-etags), and it's not published. I've tested the hash generated vs gsutil, so it should be correct unless google changes the algorithm

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/model.scala
Qi77Qi: I don't know if swagger codegen will give me circe encoder/decoder. But without swagger codegen, it's pretty easy just to generate encoder/decoder (all you need is `import io.circe.generic.auto._`), but I prefer to write encoder/decoder manually because it gives us more control, and we can bake more logic into it for validating json, and better error messages

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/model.scala
Qi77Qi: yea, you're right it doesn't...The logic was originally inside the `Decoder` below, but later I realize I could reuse it, so I just kinda temporarily put it here. I'll move it to a better place (the PR is still wip btw :))

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
Qi77Qi: there's a TODO line few lines below :)

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
Qi77Qi: Ideally some caching would be nice...I'll leave this to Matt since he's working on the 2 storagelinks endpoints

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
Qi77Qi: good point. I'll leave a TODO. I haven't had a chance to ask about what `recursive` flag does yet.

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
Qi77Qi: All exceptions are handled [here](https://github.com/DataBiosphere/welder/pull/3/files#diff-68d0c18ce140ca399b691d8fd29f6601R23) and translated to appropriate http code.

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
Qi77Qi: it'll show up in logs. Mostly the intention is we'll be able to find all logs related say one request. But later I realize it probably doesn't help that much because welder lives on user's vm

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
Qi77Qi: this is not nit :)

I was coding based on assumption here and should've left a TODO fo confirming if this is what we want

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/Crc32c.scala
rtitle: What about https://docs.oracle.com/javase/8/docs/api/java/util/zip/CRC32.html?

server/src/test/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectServiceSpec.scala
```diff
@@ -0,0 +1,63 @@
...
+import org.scalatest.FlatSpec
+
+
+import scala.concurrent.ExecutionContext.global
+
+class ObjectServiceSpec extends FlatSpec with WelderTestSuite {
+  implicit val unsafeLogger: Logger[IO] = Slf4jLogger.getLogger[IO]
+
+  val objectService = ObjectService(FakeGoogleStorageInterpreter, global, Paths.get("fakePath"))
+  "ObjectService" should "be able to localize a file" in {
```
jmthibault79: Please add these test cases:
* Localize: Google Object not found
* Delocalize (positive)
* Delocalize: Google Object not found
* Delocalize: file is outdated
* Metadata (positive for each SyncStatus)
* Metadata: No storage link found
* Metadata: Google Object not found
* Metadata: Google Object Metadata is OK but we can't retrieve the object itself

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/Crc32c.scala
Qi77Qi: Google's doc says their crc32 is different from the regular java one

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
Qi77Qi: This is correct

```@ val stream: Stream[IO, Byte] = Stream.empty
stream: Stream[IO, Byte] = Stream(..)

@ stream.compile.to[Array]
res7: IO[Array[Byte]] = Map(Bind(Delay(fs2.Stream$Compiler$$anon$5$$Lambda$1957/638580151@673e76b3), fs2.Stream$Compiler$$anon$5$$Lambda$1958/832322085@1bedc703), fs2.Stream$CompileOps$$Lambda$2045/1562251195@7c011174, 0)

@ res7.unsafeRunSync
res8: Array[Byte] = Array()

@ res8.length
res9: Int = 0```

File: server/src/test/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectServiceSpec.scala (hunk 17 above)
Comment (Qi77Qi) [reply server/src/test/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectServiceSpec.scala]: thanks @jmthibault79 for reminder.

I think lots of the failure path errors are still tbd, so I'll add doc for those first before I add tests since likely expected behavior will change

File: core/src/main/scala/org/broadinstitute/dsp/workbench/welder/Crc32c.scala (hunk 1 above)
Comment (rtitle) [reply core/src/main/scala/org/broadinstitute/dsp/workbench/welder/Crc32c.scala]: Oh missed the `c` :) ok

File: core/src/test/scala/org/broadinstitute/dsp/workbench/hamm/Generators.scala [hunk 18]
```diff
@@ -0,0 +1,18 @@
+package org.broadinstitute.dsp.workbench.welder
```
jmthibault79: A few of these are in the `hamm` dir instead of `welder`

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/Crc32c.scala
calbach: I also missed this initially - that's what threw me off. This makes sense then. I would probably add a comment somewhere in this file that mentions this is a variation of CRC32 to help future readers.

As an alternative, we could probably use MD5 instead if this presents any problems (but I think current approach is fine AFAICT for now).

geoffjentry: It's worth also making it clear that it's **crc32c** as that makes it easier to dig up information

server/src/main/resources/api-docs.yaml
Qi77Qi: what other fields are you thinking for error message response?

calbach: I don't think we need anything else currently, but other common things to return here include:
- application level error code or enum (you have specific application logic where status code alone is not sufficient to disambiguate the situation)
- internal error message or stack trace (vs public message); in general I wouldn't return that for security reasons but it can be useful if you're able to limit the circumstances where it is returned (internal IPs only, or dev-only)

Qi77Qi: Ok. I can see us potentially have application specific error code.

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/model.scala
```diff
@@ -1,9 +1,46 @@
...
+import ca.mrvisser.sealerate
+import org.broadinstitute.dsde.workbench.google2.GcsBlobName
+import org.broadinstitute.dsde.workbench.model.WorkbenchEmail
+import org.broadinstitute.dsde.workbench.model.google.GcsBucketName
+import org.http4s.Uri
 
 final case class LocalObjectPath(asString: String) extends AnyVal
+sealed abstract class SyncStatus extends Product with Serializable
+object SyncStatus {
+  // crc32 match
```
MatthewBemis: should read "crc32c match"

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/model.scala
```diff
@@ -1,9 +1,46 @@
...
+  final case object Live extends SyncStatus {
+    override def toString: String = "LIVE"
+  }
+  // crc32 mismatch
```
MatthewBemis: same comment as above

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -30,21 +44,104 @@ class ObjectService(googleStorageService: GoogleStorageService[IO]) extends Http
...
+          for {
+            bucketAndObject <- IO.fromEither(parseGsDirectory(sl.cloudStorageDirectory.renderString).leftMap(e => InternalException(e)))
+            metadata <- googleStorageService.getObjectMetadata(bucketAndObject.bucketName, bucketAndObject.blobName, Some(traceId)).compile.lastOrError
+            result <- metadata match {
+                        case google2.GetMetadataResponse.NotFound =>
+                          IO.pure(MetadataResponse.RemoteNotFound(sl))
+                        case google2.GetMetadataResponse.Metadata(crc32, userDefinedMetadata) =>
+                          for {
+                            isLinked <- IO.fromEither(userDefinedMetadata.get("isLinked").fold[Either[Throwable, Boolean]](Right(false))(x => Either.catchNonFatal(x.toBoolean)))
+                            fileBody <- googleStorageService.getObject(bucketAndObject.bucketName, bucketAndObject.blobName, Some(traceId)).compile.to[Array]
```
calbach: The goal is to compare the crc32c from the GCS metadata against the crc32c of the local file. This appears to be recomputing the crc32c of the remote object (which would effectively just be verifying that GCS set the metadata value properly)

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -30,21 +44,104 @@ class ObjectService(googleStorageService: GoogleStorageService[IO]) extends Http
...
+        sl =>
+          for {
+            bucketAndObject <- IO.fromEither(parseGsDirectory(sl.cloudStorageDirectory.renderString).leftMap(e => InternalException(e)))
+            metadata <- googleStorageService.getObjectMetadata(bucketAndObject.bucketName, bucketAndObject.blobName, Some(traceId)).compile.lastOrError
+            result <- metadata match {
+                        case google2.GetMetadataResponse.NotFound =>
+                          IO.pure(MetadataResponse.RemoteNotFound(sl))
+                        case google2.GetMetadataResponse.Metadata(crc32, userDefinedMetadata) =>
+                          for {
+                            isLinked <- IO.fromEither(userDefinedMetadata.get("isLinked").fold[Either[Throwable, Boolean]](Right(false))(x => Either.catchNonFatal(x.toBoolean)))
```
calbach: I assume userDefinedMetadata is coming from GCS? "isLinked" isn't something I was proposing we would have on the GCS object metadata. This should be purely be determined by whether or not storageLink was found

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -30,21 +44,104 @@ class ObjectService(googleStorageService: GoogleStorageService[IO]) extends Http
...
+
+    res.compile.drain
+  }
 
-  def checkMeta(req: GetMetadataRequest): IO[Unit] = ???
+  def checkMetadata(req: GetMetadataRequest): IO[MetadataResponse] =
+    for {
+      traceId <- IO(TraceId(randomUUID()))
+      storageLinks <- readJsonFileToA[IO, List[StorageLink]](pathToStorageLinks).compile.lastOrError
+      storageLink = storageLinks.find(_.localBaseDirectory == req.localObjectPath) //TODO: handle recursive flag
```
calbach: IF we're in execution mode, it should also return the storageLink

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/model.scala
Qi77Qi: updated

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -30,21 +44,104 @@ class ObjectService(googleStorageService: GoogleStorageService[IO]) extends Http
...
+          } yield ()
+      }
+    } yield ()
+  }
+
+  private def delocalize(localObjectPath: LocalObjectPath, gsPath: Uri, traceId: TraceId): IO[Unit] = io.file.readAll[IO](Paths.get(localObjectPath.asString), blockingEc, 4096).compile.to[Array].flatMap {
+    body =>
+      for {
+        bucketNameAndObjectName <- IO.fromEither(parseGsDirectory(gsPath.renderString).leftMap(InternalException))
+        _ <- googleStorageService.storeObject(bucketNameAndObjectName.bucketName, bucketNameAndObjectName.blobName, body, gcpObjectType, Some(traceId))
```
calbach: Per the design doc, please pass the x-goog-if-generation-match header with the current known generation version. This could potentially be a follow-up change.

It would also be better to cache this metadata in memory so you don't need to refetch the GCS metadata on each delocalize call (this will slowdown the application).

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/model.scala
```diff
@@ -1,9 +1,46 @@
...
+  }
+  // deleted in gcs. (object exists in storagelinks config file but not in in gcs)
+  final case object RemoteNotFound extends SyncStatus {
+    override def toString: String = "REMOTE_NOT_FOUND"
+  }
+
+  val stringToSyncStatus: Set[SyncStatus] = sealerate.values[SyncStatus]
+}
+final case class BucketNameAndObjectName(bucketName: GcsBucketName, blobName: GcsBlobName) {
+  override def toString: String = s"${bucketName.value}/${blobName.value}"
```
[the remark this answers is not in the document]
Qi77Qi: I didn't mean to have `BucketNameAndObjectName` represent a `gs` path..it just happens to come from gs path from user input, but the data type itself is different from a gs path.

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/package.scala
```diff
@@ -0,0 +1,34 @@
...
+import io.circe.Decoder
+import org.broadinstitute.dsde.workbench.google2.GcsBlobName
+import org.broadinstitute.dsde.workbench.model.google.GcsBucketName
+import org.typelevel.jawn.AsyncParser
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.ExecutionContext.global
+
+package object welder {
+  def readJsonFileToA[F[_]: Sync: ContextShift: Concurrent, A: Decoder](path: Path, blockingExecutionContext: Option[ExecutionContext] = None): Stream[F, A] = {
```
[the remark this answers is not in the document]
Qi77Qi: It's a common thing to use to represent universally quantified variable. Here, `A` can be any type. So as long as there's a `Decoder[A]` instance in scope, this function will be able to read bytes from a file and convert it into a target `A` type

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/package.scala
```diff
@@ -0,0 +1,34 @@
...
+    fs2.io.file.readAll[F](path, blockingExecutionContext.getOrElse(global), 4096)
+      .through(fs2.text.utf8Decode)
+      .through(_root_.io.circe.fs2.stringParser(AsyncParser.SingleValue))
+      .through(decoder)
+  }
+
+  def parseGsDirectory(str: String): Either[String, BucketNameAndObjectName] = for {
+    parsed <- Either.catchNonFatal(str.split("/")).leftMap(_.getMessage)
```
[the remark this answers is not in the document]
Qi77Qi: sure..I'll add that

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
Qi77Qi: we're only computing the crc32c for local file. `crc32` variable is from calling gcs.

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
Qi77Qi: > I assume userDefinedMetadata is coming from GCS?
Yes.
And thanks for clarification

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
Qi77Qi: I need more context about this comment

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
Qi77Qi: where does welder get generation version from?
We're using google's java lib (grpc), so it wouldn't be a header, but I think we do have option to pass it a generation variable

server/src/main/resources/api-docs.yaml
```diff
@@ -0,0 +1,128 @@
...
+##########################################################################################
+paths:
+  '/objects/metadata':
+    get:
+      summary: 
+      responses:
+        '200':
+          schema:
+            oneof:
+              - $ref: '#/definitions/MetadataRemoteNotFound'
```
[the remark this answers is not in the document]
Qi77Qi: updated

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
calbach: It should be returned as part of the GCS object metadata. At the transport level, it's a header - but specifically for the Google Java Lib you will probably need `BlobTargetOption.generationMatch(XXX)`

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -30,21 +43,102 @@ class ObjectService(googleStorageService: GoogleStorageService[IO]) extends Http
...
+    }.parJoin(10)
+
+    res.compile.drain
+  }
+
+  def checkMetadata(req: GetMetadataRequest): IO[MetadataResponse] =
+    for {
+      traceId <- IO(TraceId(randomUUID()))
+      storageLinks <- readJsonFileToA[IO, List[StorageLink]](pathToStorageLinks).compile.lastOrError
+      storageLink = storageLinks.find(_.localBaseDirectory == req.localObjectPath) //TODO: handle recursive flag
```
calbach: I expect `localObjectPath` to be something like "workspace-dir/notebook1.ipynb". I expect `localBaseDirectory` to be something like "workspace-dir". so this predicate should be matching the basedir of the file instead.

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
calbach: I think this would probably be a follow-up PR, but I added some examples to the design doc appendix: https://docs.google.com/document/d/1rTq3DwsB2h7l_9f2I5pcpa9Z-PJJkMBw7APBqYs1NW8/edit#heading=h.mchwz84nu1kh

After looking at the examples I'm going to propose some tweaks to this aspect of the API in the doc, curious about your thoughts on this (see comment).

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -30,21 +43,102 @@ class ObjectService(googleStorageService: GoogleStorageService[IO]) extends Http
       } yield resp
   }
 
-  def localize(req: Localize): IO[Unit] = ???
+  def localize(req: Localize): IO[Unit] = {
+    val res = Stream.emits(req.entries).map {
+      entry =>
+        googleStorageService.getObject(entry.bucketNameAndObjectName.bucketName, entry.bucketNameAndObjectName.blobName, None) //get file from google. TODO: pass generation when storing object
```
calbach: This TODO is not necessary in the current proposed design. The generation would only come from checkMeta to keep the Leo VM as stateless as possible on file syncing.

calbach: Per design doc, this should also support b64 encoded files to match existing localize extension behavior. Definitely should just be a TODO though.

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
Qi77Qi: This is for storing object to gcs, so we don't have metadata for the object yet. Or am I missing something? Speaking of which, is /safeDelocalize supposed to adding metadata like lastEditedBy etc as well?

calbach: If the object doesn't exist, we should pass generation of 0. This method will also be used for subsequent updates to the file from the Jupyter UI, in which case we want to send the previous known generation ID.

Good catch - I didn't call it out here, but yes it should also be adding the lastEdited/lastEditedBy metadata.

Qi77Qi:
I'm going to merge this PR for now...and address some unaddressed comments in a followup PR after clarification just so we can move this long

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
Qi77Qi: do you mean before we delocalize, we'll always do a getMetadata call to get generation, increment that, and use that as the new generation number?

calbach: Yes - except I would use the cached value here. There's no reason to refetch the metadata before delocalizing (unless we've never fetched the metadata before).

The generation number is just a value returned by GCS - we would need to refetch it at some point and should not attempt to increment it manually. More here: https://cloud.google.com/storage/docs/generations-preconditions

calbach: (probably it would look like: 1. update 2. invalidate metadata cache for that file)

PR #6, opened 2019-05-23
Address TODOs

Qi77Qi:
- support data uri
- add more unit tests (will add more)
- pass generation when delocalizing
   - we update gcs metadata cache(including generation) every time we localize a file

TODO:
- execution mode (depends on Matt's change)
- unit tests

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsp/workbench/welder/model.scala
+++ b/core/src/main/scala/org/broadinstitute/dsp/workbench/welder/model.scala
@@ -11,6 +11,7 @@ import cats.effect.IO
 import org.broadinstitute.dsde.workbench.google2.{Crc32, GcsBlobName}
 import org.broadinstitute.dsde.workbench.model.WorkbenchEmail
 import org.broadinstitute.dsde.workbench.model.google.GcsBucketName
+import org.broadinstitute.dsp.workbench.welder.LocalBasePath.{LocalBaseDirectoryPath, LocalSafeBaseDirectoryPath}
 import org.broadinstitute.dsp.workbench.welder.SourceUri.{DataUri, GsPath}
 import org.http4s.Uri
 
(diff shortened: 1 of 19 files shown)


server/src/test/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectServiceSpec.scala
```diff
@@ -2,31 +2,33 @@ package org.broadinstitute.dsp.workbench.welder
...
-import org.scalatestplus.scalacheck.ScalaCheckPropertyChecks
-import org.broadinstitute.dsp.workbench.welder.Generators.{arbGcsBlobName, arbGcsBucketName}
 
 import scala.concurrent.ExecutionContext.global
 
-class ObjectServiceSpec extends FlatSpec with ScalaCheckPropertyChecks with WelderTestSuite {
+class ObjectServiceSpec extends FlatSpec with WelderTestSuite {
   implicit val unsafeLogger: Logger[IO] = Slf4jLogger.getLogger[IO]
-  val objectService = ObjectService(FakeGoogleStorageInterpreter, global, Paths.get("fakePath"))
+  val storageLinksCache = Ref.of[IO, Map[Path, StorageLink]](Map.empty).unsafeRunSync()
```
MatthewBemis: can this be Ref.of[StorageLinksCache] now?

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/model.scala
```diff
@@ -28,10 +31,27 @@ object SyncStatus {
...
 
+sealed abstract class SourceUri
+object SourceUri {
+  final case class DataUri(data: Array[Byte]) extends SourceUri
+  final case class GsPath(bucketName: GcsBucketName, blobName: GcsBlobName) extends SourceUri {
+    override def toString: String = s"gs://${bucketName.value}/${blobName.value}"
+  }
+}
+
+final case class StorageLink(localBaseDirectory: Path, cloudStorageDirectory: GsPath, pattern: String, recursive: Boolean)
```
MatthewBemis: should we also move `StorageLinks` and the encoders/decoders here?

server/src/test/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectServiceSpec.scala
Qi77Qi: Not really. `StorageLinksCache` is the whole `Ref`....`type StorageLinksCache = Ref[IO, Map[Path, StorageLink]]`

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/model.scala
Qi77Qi: sounds good...

server/src/test/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectServiceSpec.scala
MatthewBemis: ah oops

PR #7, opened 2019-05-23
Store storageLinks in memory, write to file upon welder shutdown

MatthewBemis:


Qi77Qi:
can you add api doc as well

MatthewBemis:
merging. will add tests and docs in followup PR

PR #9, opened 2019-05-28
Add unit tests and Swagger docs for /storageLinks endpoints

MatthewBemis:


server/src/test/scala/org/broadinstitute/dsp/workbench/welder/server/StorageLinksServiceSpec.scala
```diff
@@ -0,0 +1,92 @@
...
+    assert(addResult equals linkToAdd)
+  }
+
+  it should "not create duplicate storage links" in {
+    val storageLinks = Ref.of[IO, Map[Path, StorageLink]](Map.empty)
+    val storageLinksService = StorageLinksService(storageLinks.unsafeRunSync())
+
+    val linkToAdd = StorageLink(Paths.get("/foo"), Uri.unsafeFromString("gs://foo/bar/baz.zip"), ".zip", true)
+
+    storageLinksService.createStorageLink(linkToAdd).unsafeRunSync()
```
Qi77Qi: nit, I'll use for comprehension and use `unsafeRunSync()` only once at the end of each test

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/Main.scala
```diff
@@ -25,7 +25,10 @@ object Main extends IOApp {
     val app: Stream[IO, Unit] = for {
       appConfig <- Stream.fromEither[IO](Config.appConfig)
       blockingEc <- Stream.resource[IO, ExecutionContext](ExecutionContexts.fixedThreadPool(255))
-      storageLinks <- readJsonFileToA[IO, StorageLinks](appConfig.pathToStorageLinksJson).map(sl => sl.storageLinks.map(s => s.localBaseDirectory -> s).toMap)
+      storageLinks <- readJsonFileToA[IO, StorageLinks](appConfig.pathToStorageLinksJson).map(sl => sl.storageLinks.map(s => s.localBaseDirectory -> s).toMap).handleErrorWith { error =>
+        Stream.eval(Logger[IO].error(error)(s"Failed to load storage links: ${error.getMessage}"))
```
Qi77Qi: I think you'll need to `Stream.eval(Logger[IO].error(error)(s"Failed to load storage links: ${error.getMessage}")) *>  Stream.resource[IO, Map[Path, StorageLink]](Resource.pure(Map[Path, StorageLink]()))`

The first log won't actually get executed.

For instance:
```

@ def foo: IO[Unit] = {
  IO(println(1))
  IO(println(2))
  }
defined function foo

@ foo.unsafeRunSync
2
```

PR #10, opened 2019-06-03
Diagram and unit tests

Qi77Qi:
- Add unit tests for /checkMeta, /safeDelocalize
- persist metadata cache
- Add diagram flow


server/src/main/resources/UI_Interaction.puml
```diff
@@ -0,0 +1,54 @@
+@startuml
+participant TerraUI
+box "Leo VM"
+    participant JupyterExtension
```
calbach: What is JupyterExtension vs JupyterAutoSaveExtension? Maybe Jupyter UI Extension and Jupyter Server Extension would be clearer (I assume that is the distinction here)

server/src/main/resources/UI_Interaction.puml
```diff
@@ -0,0 +1,54 @@
...
+        Welder -> JupyterExtension: Locked by someone else
+        note left: prompt user to choose 1. make a new copy  2. go to safe mode with current file 3. go to SAFE mode with latest remote file
+    end
+    loop how often?
+        JupyterAutoSaveExtension -> Welder: /safeDelocalize
+    end
+else 200 SyncStatus.SAFE
+    Welder -> JupyterExtension: SAFE mode
+else 412 StorageLink not found
+    return storagelink not found for this file. UI should do /postStorageLink
```
calbach: Which UI should do a POST? This is a valid situation - the UI extension should do nothing here.

server/src/main/resources/UI_Interaction.puml
```diff
@@ -0,0 +1,54 @@
...
+
+    alt 200 successfully acquire Lock
+        Welder -> JupyterExtension: No Lock or locked by current user
+        loop every 30 seconds?
+            JupyterExtension -> Welder ++ #F4D03F: /checkMeta
+            alt SyncStatus.LIVE
+                Welder -> JupyterExtension: file is live
+            else SyncStatus.REMOTE_NOT_FOUND
+                Welder -> JupyterExtension: remote file deleted
+                note left: Do we assume user will just want to /safeDelocalize?
```
calbach: The UI extension could offer this as an option. We shouldn't assume this though and automatically save - we would want to let the user know they are desycnhronized.

server/src/main/resources/UI_Interaction.puml
```diff
@@ -0,0 +1,54 @@
...
+            else SyncStatus.DESYNCHRONIZED
+                return file is out of sync
+                note left: prompt user to choose 1. make a new copy 2. go to SAFE mode with current file 3. go to SAFE mode with latest remote file
+            end
+        end
+    else lockedBy someone else
+        Welder -> JupyterExtension: Locked by someone else
+        note left: prompt user to choose 1. make a new copy  2. go to safe mode with current file 3. go to SAFE mode with latest remote file
```
calbach: The 2 vs 3 distinction is interesting, but may be too subtle to try and explain in a modal. Discussion is continuing with Marty/Adrian on what the UX should be here

server/src/main/resources/UI_Interaction.puml
```diff
@@ -0,0 +1,54 @@
...
+    end
+    loop how often?
```
calbach: This is not a loop, this is a hook into the server-side Jupyter file-system handler. Whenever a file gets created/deleted/moved/saved via the Jupyter API, we would do a safeDelocalize.

server/src/main/resources/UI_Interaction.puml
Qi77Qi: it'll only be valid if the file is not being monitored by welder. But if the file is not monitored by welder, why would ui extension call `/checkMeta` for this file?

calbach: This endpoint is how the UI extension determines whether this file is being monitored by welder.

server/src/main/resources/UI_Interaction.puml
Qi77Qi: If `SyncStatus.REMOTE_NOT_FOUND` happens in `/safeDelocalize`, shall we do nothing then?

calbach: We should delocalize. Can set a generation precondition which matches this (I think 0 for creating new files).

Qi77Qi: What's the point of precondition in the case of `RMOTE_NOT_FOUND`?

calbach: To avoid two users writing to the same location simultaneously. Very unlikely happen, but doesn't hurt to set it.

jdcanas:
Can the /acquireLock endpoint throw a 409 (or some other error code) if you attempt to grab the lock on a locked notebook?

Qi77Qi:
@jdcanas , I think it would... @MatthewBemis is working on that

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -82,103 +82,112 @@ class ObjectService(
     res.compile.drain
   }
 
-  private def retrieveGcsMetadata(bucketName: GcsBucketName, blobName: GcsBlobName, traceId: TraceId): IO[Option[GcsMetadata]] = for {
+  private def retrieveGcsMetadata(localPath: java.nio.file.Path, bucketName: GcsBucketName, blobName: GcsBlobName, traceId: TraceId): IO[Option[GcsMetadata]] = for {
```
jdcanas: the number of arguments and length of the method seems to indicate a violation of the single responsibility principle (in this case both the processing and packaging of the data). Consider making a helper method

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -82,103 +82,112 @@ class ObjectService(
...
-              currentTime <- timer.clock.realTime(TimeUnit.MILLISECONDS)
-              lastLock <- expiresAt.flatTraverse[IO, WorkbenchEmail] { ea =>
-                if (currentTime > ea.toEpochMilli)
-                  IO.pure(none[WorkbenchEmail])
-                else
-                  IO.pure(lastLockedBy)
-              }
-            } yield MetadataResponse.EditMode(syncStatus, lastLock, isSafeMode, generation, sl)
+      storageLink <- findStorageLink(req.localObjectPath)
+      res <- storageLink.fold[IO[MetadataResponse]](IO.raiseError(StorageLinkNotFoundException(s"No storage link found for ${req.localObjectPath}"))) { pair =>
```
jdcanas: this method is heavily nested (i think at one point there are 6 layers of complexity, but counting it is a bit difficult in git). considering refactoring some logic to a helper function for readability.

server/src/main/resources/api-docs.yaml
```diff
@@ -138,6 +152,8 @@ definitions:
       - LIVE
       - DESYNCHRONIZED
       - REMOTE_NOT_FOUND
+      - REMOTE_CHANGED
+      - LOCAL_CHANGED
```
calbach: I don't think this case can be reliably detected in the current design. I would recommend removing (and just returning the more general "DESYNCHRONIZED" here).

server/src/main/resources/api-docs.yaml
```diff
@@ -138,6 +152,8 @@ definitions:
       - LIVE
       - DESYNCHRONIZED
```
calbach: I would document the relationship between this and the other values. Possibly suffix of _OTHER or _UNKNOWN

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -82,103 +82,112 @@ class ObjectService(
...
-              else
-                IO.raiseError(FileOutOfSyncException("file is updated in google storage by someone else"))
-            case (_, None) => // if file is deleted in remote, we'll just delocalize the file
-              delocalize(req.localObjectPath, gsPath, traceId, 0L)
-            case (None, Some(now)) =>
-              IO.raiseError(new Exception("Can't determine if the file user started off of is up-to-date"))
-          }
-        } yield ()
+        if(isSafeMode)
+          IO.unit
```
calbach: This should probably be the same behavior as when there is no storage link, i.e. 412

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -82,103 +82,112 @@ class ObjectService(
...
+        else {
+          for {
+            previousMeta <- metadataCache.get.map(_.get(req.localObjectPath))
```
calbach: Ideally this would populate on a cache miss, rather than assuming no file in GCS.

server/src/main/resources/UI_Interaction.puml
```diff
@@ -0,0 +1,60 @@
...
+            else SyncStatus.REMOTE_CHANGED
+               return: remote copy in GCS has changed
+                note left: prompt user to choose 1. make a new copy 2. go to SAFE mode with current file 3. go to SAFE mode with latest remote file
+            end
+        end
+    else lockedBy someone else
+        Welder -> JupyterUIExtension: Locked by someone else
+        note left: prompt user to choose 1. make a new copy  2. go to safe mode with current file 3. go to SAFE mode with latest remote file
+    end
+    group autosave (jupyter built-in feature), whenever a file gets created/deleted/moved/saved via the Jupyter API
```
calbach: "autosave" has gotten misused a lot. When I use this term, I am just referring to the "autosave" function of the Jupyter UI which is a typical IDE behavior of automatically saving your unsaved changes - that's it. It's pretty much entirely unrelated to the Welder work.

The thing that interfaces with Welder is the Jupyter server extension "jupyter_delocalize.py", which is implemented via a custom ContentsManager (WelderContentsManager - WIP).

project/Settings.scala
```diff
@@ -113,12 +113,5 @@ object Settings {
     )
   )
 
-  lazy val newrelicSettings = List(
```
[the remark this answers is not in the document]
Qi77Qi: leaving it out for now since welder is a service deployed on user's cluster. It's a bit more complicated enabling NR in this case. (we need a NR license key on the vm, which we were talking about potentially giving user root access)

server/src/main/resources/api-docs.yaml
Qi77Qi: wouldn't it be useful to tell client which case it is when we can?

server/src/main/resources/UI_Interaction.puml
Qi77Qi: is `jupyter_delocalize.py` still triggered by `autosave` function? or how is `jupyter_delocalize.py` gets triggered?

server/src/main/resources/api-docs.yaml
calbach: Theoretically yes, though practically speaking I think it's probably a minor value add (at most probably a different error message in the MVP).

Keeping metadata state around on disk and trying to make decisions about this sounds like a non-trivial deviation from the original proposed design. Adding a section to the design doc describing how this works might be an easier way to continue this discussion.

server/src/main/resources/UI_Interaction.puml
calbach: `jupyter_delocalize.py` hooks into the Jupyter server model. Any save, delete, file creation, file rename, etc goes through this interface. In our case, the most important hook is overriding the `FileContentsManager.save()` function; this gets called on any API save, which gets called in many situations, e.g. explicit save in Jupyter UI, "save as", new file creation, periodic UI autosaves, etc...

server/src/main/resources/api-docs.yaml
Qi77Qi: What do you mean the other values? do you mean `REMOTE_CHANGED` and `LOCAL_CHANGED`?

server/src/main/resources/api-docs.yaml
Qi77Qi: We make decision around generation regardless. The only deviation is the same data will be persisted on disk the same way storagelinks get persisted on disk

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
Qi77Qi: What metadata would you populate the cache with in this case? (We can't just make a google call and get the metadata because it wouldn't be the metadata for the local version. Gcs metadata cache is populated whenever a file is localized)

Qi77Qi:
I'm going to merge this and address comments in separate PRs if needed

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
calbach: This ties into the other discussion, my comment doesn't make sense if you're planning to precache metadata for every localized file.

PR #13, opened 2019-06-11
refactor and add delete api

Qi77Qi:
* better json decoding error in api response
* a few bug fixes
* Add delete api
* Make server port configurable

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsp/workbench/welder/package.scala
+++ b/core/src/main/scala/org/broadinstitute/dsp/workbench/welder/package.scala
@@ -39,18 +39,17 @@ package object welder {
   } yield GsPath(GcsBucketName(bucketName), GcsBlobName(objectName))
 
   // base directory example: “workspaces/ws1”
-  def getLocalBaseDirectory(localPath: Path): Either[String, Path] = {
-    for {
-      prefix <- Either.catchNonFatal(localPath.getName(0)).leftMap(_ => s"no valid prefix found for $localPath")
-      workspaceName <- Either.catchNonFatal(localPath.getName(1)).leftMap(_ => s"no workspace name found for $localPath")
-    } yield localPath.subpath(0, 2)
+  def getAllPosssibleBaseDirectory(localPath: Path): List[Path] = {
+    ((localPath.getNameCount - 1).to(1, -1)).map(
+      x => localPath.subpath(0, x)
+    ).toList
   }
 
-  def getFullBlobName(localPath: Path, blobPath: BlobPath): Either[String, GcsBlobName] =
-    for {
-      prefix <- getLocalBaseDirectory(localPath)
-      subPath = prefix.relativize(localPath)
-    } yield GcsBlobName(blobPath.asString + "/" + subPath.toString)
+  def getFullBlobName(basePath: Path, localPath: Path, blobPath: BlobPath): GcsBlobName = {
+    println(s"basePath $basePath; $localPath")
+      val subPath = basePath.relativize(localPath)
+      GcsBlobName(blobPath.asString + "/" + subPath.toString)
+    }
 
   val base64Decoder = Base64.getDecoder()
   def base64DecoderPipe[F[_]: Sync]: Pipe[F, String, Byte] = in => {
@@ -58,7 +57,7 @@ package object welder {
       .flatMap(bytes => Stream.emits(bytes).covary[F])
   }
 
-  type StorageLinksCache = Ref[IO, Map[LocalDirectory, StorageLink]]
+  type StorageLinksCache = Ref[IO, Map[Path, StorageLink]]
   type MetadataCache = Ref[IO, Map[Path, GcsMetadata]]
 
   val gcpObjectType = "text/plain"
--- a/core/src/main/scala/org/broadinstitute/dsp/workbench/welder/package.scala
+++ b/core/src/main/scala/org/broadinstitute/dsp/workbench/welder/package.scala
@@ -27,20 +27,28 @@ package object welder {
   }
 
   val gsDirectoryReg = "gs:\\/\\/.*".r
-  def parseGsPath(str: String): Either[String, GsPath] = for {
-    _ <- gsDirectoryReg.findPrefixOf(str).toRight("gs directory has to be prefixed with gs://")
+
+  def validateGsPrefix(str: String): Either[String, Unit] = gsDirectoryReg.findPrefixOf(str).void.toRight("gs directory has to be prefixed with gs://")
+
+  def parseBucketName(str: String): Either[String, GcsBucketName] = for {
+    _ <- validateGsPrefix(str)
     parsed <- Either.catchNonFatal(str.split("/")).leftMap(_.getMessage)
     bucketName <- Either.catchNonFatal(parsed(2))
       .leftMap(_ => s"failed to parse bucket name")
       .ensure("bucketName can't be empty")(s => s.nonEmpty)
-    objectName <- Either.catchNonFatal(parsed.drop(3).mkString("/"))
+  } yield GcsBucketName(bucketName)
+
+  def parseGsPath(str: String): Either[String, GsPath] = for {
+    bucketName <- parseBucketName(str)
+    length = s"gs://${bucketName.value}/".length
+    objectName <- Either.catchNonFatal(str.drop(length))
       .leftMap(_ => s"failed to parse object name")
       .ensure("objectName can't be empty")(s => s.nonEmpty)
-  } yield GsPath(GcsBucketName(bucketName), GcsBlobName(objectName))
+  } yield GsPath(bucketName, GcsBlobName(objectName))
 
   def getPosssibleBaseDirectory(localPath: Path): List[Path] = {
     ((localPath.getNameCount - 1).to(1, -1)).map(
-      x => localPath.subpath(0, x)
+      index => localPath.subpath(0, index)
     ).toList
   }
 
--- a/core/src/main/scala/org/broadinstitute/dsp/workbench/welder/package.scala
+++ b/core/src/main/scala/org/broadinstitute/dsp/workbench/welder/package.scala
@@ -38,15 +38,13 @@ package object welder {
       .ensure("objectName can't be empty")(s => s.nonEmpty)
   } yield GsPath(GcsBucketName(bucketName), GcsBlobName(objectName))
 
-  // base directory example: “workspaces/ws1”
-  def getAllPosssibleBaseDirectory(localPath: Path): List[Path] = {
+  def getPosssibleBaseDirectory(localPath: Path): List[Path] = {
     ((localPath.getNameCount - 1).to(1, -1)).map(
       x => localPath.subpath(0, x)
     ).toList
   }
 
   def getFullBlobName(basePath: Path, localPath: Path, blobPath: BlobPath): GcsBlobName = {
-    println(s"basePath $basePath; $localPath")
       val subPath = basePath.relativize(localPath)
       GcsBlobName(blobPath.asString + "/" + subPath.toString)
     }
--- a/server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/Main.scala
+++ b/server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/Main.scala
@@ -33,7 +33,7 @@ object Main extends IOApp {
       storageLinksService = StorageLinksService(storageLinksCache)
       objectServiceConfig = ObjectServiceConfig(appConfig.workingDirectory, appConfig.currentUser, appConfig.lockExpiration)
       objectService = ObjectService(objectServiceConfig, googleStorageService, blockingEc, storageLinksCache, metadataCache)
-      server <- BlazeServerBuilder[IO].bindHttp(8080, "0.0.0.0").withHttpApp(WelderApp(objectService, storageLinksService).service).serve
+      server <- BlazeServerBuilder[IO].bindHttp(appConfig.serverPort, "0.0.0.0").withHttpApp(WelderApp(objectService, storageLinksService).service).serve
     } yield ()
 
     app.handleErrorWith {
--- a/server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/Main.scala
+++ b/server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/Main.scala
@@ -27,7 +27,7 @@ object Main extends IOApp {
       blockingEc <- Stream.resource[IO, ExecutionContext](ExecutionContexts.fixedThreadPool(255))
       implicit0(it: Linebacker[IO])  = Linebacker.fromExecutionContext[IO](blockingEc)
       appConfig <- Stream.fromEither[IO](Config.appConfig)
-      storageLinksCache <- cachedResource[LocalDirectory, StorageLink](appConfig.pathToStorageLinksJson, blockingEc, storageLink => List(storageLink.localBaseDirectory -> storageLink, storageLink.localSafeModeBaseDirectory -> storageLink))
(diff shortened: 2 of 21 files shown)


server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/Main.scala
```diff
@@ -17,27 +18,28 @@ import org.broadinstitute.dsp.workbench.welder.JsonCodec._
...
-import scala.concurrent.duration._
 
 object Main extends IOApp {
   override def run(args: List[String]): IO[ExitCode] = {
     implicit val unsafeLogger = Slf4jLogger.getLogger[IO]
 
     val app: Stream[IO, Unit] = for {
+      blockingEc <- Stream.resource[IO, ExecutionContext](ExecutionContexts.fixedThreadPool(255))
+      implicit0(it: Linebacker[IO])  = Linebacker.fromExecutionContext[IO](blockingEc)
       appConfig <- Stream.fromEither[IO](Config.appConfig)
```
calbach: Does this handle the case I hit today, where `appConfig` returned an error?

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -180,18 +193,18 @@ class ObjectService(
...
 
-  private def findStorageLink(localPath: java.nio.file.Path): IO[Option[(Boolean, StorageLink)]] = for {
-    storageLinks <- storageLinksCache.get
-    baseDirectory <- IO.fromEither(getLocalBaseDirectory(localPath).leftMap(s => BadRequestException(s)))
-  } yield {
-    storageLinks.find(x => x._1.path == baseDirectory).map {
+  private def findStorageLink[A](localPath: java.nio.file.Path, safeModeResponse: StorageLink => Kleisli[IO, TraceId, A], editModeResponse: StorageLink => Kleisli[IO, TraceId, A]): Kleisli[IO, TraceId, A] = for {
+    storageLinks <- Kleisli.liftF(storageLinksCache.get)
+    baseDirectory <- Kleisli.liftF(IO.fromEither(getLocalBaseDirectory(localPath).leftMap(s => BadRequestException(s))))
+    storagelink = storageLinks.find(x => x._1.path == baseDirectory)
```
calbach: What is `path` here? Is this a real variable? Shouldn't this be comparing against the localBaseDirectoryPath?

FWIW I can't seem to produce a storageLink hit locally - I tried patching this and still see the same issue:

```
curl localhost:8080/storageLinks                                                                   
{"storageLinks":[{"localBaseDirectory":"foo","localSafeModeBaseDirectory":"foo-safe","cloudStorageDirectory":"gs://fc-0bf6534f-a9cb-4d58-b311-c6a1c65a359f/welder-test","pattern":""}]}

curl -X POST localhost:8080/objects/metadata -d '{"localPath": "foo/bar.ipynb"}'
{"errorMessage":"No storage link found for foo/bar.ipynb","errorCode":1}
```

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/Main.scala
Qi77Qi: it's not this change that addresses the issue. Reason why the app hangs previously is https://github.com/DataBiosphere/welder/pull/13/files#diff-5b4ceaa24c625c1442c692ab05b7b26dR41..So previous, we recover from any error that happens during initialization, and log the error, but then never throw the exception. Hence, it reaches `IO.never` as usual, which hangs the app.

The change will throw any exception during initialization instead.

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
Qi77Qi: [this](https://github.com/DataBiosphere/welder/blob/master/core/src/main/scala/org/broadinstitute/dsp/workbench/welder/model.scala#L62) is how baseDirectory is defined. In storagelinks, we store both `localBaseDirectory`, and `localSafeModeBaseDirectory`, and when we find a match, we then determine which one it is based on its type.

> What is path here? Is this a real variable? Shouldn't this be comparing against the localBaseDirectoryPath?

Not sure what you mean by "real variable". Its type is java.nio.file.Path, and it can be either `localBaseDirectoryPath` or `localSafeBaseDirectoryPath`. We're comparing it against calculated "baseDirectory" (which is also java,nio.file.Path) from "localPath" in user request.

calbach: OK, I can't fully trace the caching + `LocalDirectory` so I'll assume that part works.

This line I think is a problem:
https://github.com/DataBiosphere/welder/blob/70d3d77060c772af1510170f9ee2312e3fe326e0/core/src/main/scala/org/broadinstitute/dsp/workbench/welder/package.scala#L46

My example requests I gave of "/workspaces/ws1" were just an example of how the applications might structure the directories. Welder **should not make any assumptions about these path structures or any assumptions about workspaces**. It should just accept any base directory path. All paths should be relative to the Jupyter user home directory (same as the Welder volume).

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/package.scala
```diff
@@ -39,26 +39,25 @@ package object welder {
   } yield GsPath(GcsBucketName(bucketName), GcsBlobName(objectName))
 
   // base directory example: “workspaces/ws1”
-  def getLocalBaseDirectory(localPath: Path): Either[String, Path] = {
-    for {
-      prefix <- Either.catchNonFatal(localPath.getName(0)).leftMap(_ => s"no valid prefix found for $localPath")
-      workspaceName <- Either.catchNonFatal(localPath.getName(1)).leftMap(_ => s"no workspace name found for $localPath")
-    } yield localPath.subpath(0, 2)
+  def getAllPosssibleBaseDirectory(localPath: Path): List[Path] = {
```
calbach: possible

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/package.scala
```diff
@@ -39,26 +39,25 @@ package object welder {
...
+    ).toList
   }
 
-  def getFullBlobName(localPath: Path, blobPath: BlobPath): Either[String, GcsBlobName] =
-    for {
-      prefix <- getLocalBaseDirectory(localPath)
-      subPath = prefix.relativize(localPath)
-    } yield GcsBlobName(blobPath.asString + "/" + subPath.toString)
+  def getFullBlobName(basePath: Path, localPath: Path, blobPath: BlobPath): GcsBlobName = {
+    println(s"basePath $basePath; $localPath")
```
calbach: rm debug line

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -100,73 +103,92 @@ class ObjectService(
...
+              val localAbsolutePath = config.workingDirectory.resolve(req.localObjectPath)
+              for {
+                fileBody <- io.file.readAll[IO](localAbsolutePath, blockingEc, 4096).compile.to[Array]
+                calculatedCrc32c = Crc32c.calculateCrc32c(fileBody)
+                syncStatus <- if (calculatedCrc32c == crc32c) IO.pure(SyncStatus.Live) else {
+                  for {
+                    cachedGeneration <- metadataCache.get.map(_.get(req.localObjectPath))
+                    status <- cachedGeneration match {
+                      case Some(cachedGen) => if(cachedGen.generation == generation) IO.pure(SyncStatus.LocalChanged) else IO.pure(SyncStatus.RemoteChanged)
+                      case None => IO.pure(SyncStatus.OutOfSync)
```
calbach: What is OutOfSync? Is this the same as Desynchronized? If so - why a different name?

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -100,73 +103,92 @@ class ObjectService(
...
+            case None =>
+              IO(MetadataResponse.RemoteNotFound(pair.storageLink))
+            case Some(GcsMetadata(_, lastLockedBy, expiresAt, crc32c, generation)) =>
+              val localAbsolutePath = config.workingDirectory.resolve(req.localObjectPath)
+              for {
+                fileBody <- io.file.readAll[IO](localAbsolutePath, blockingEc, 4096).compile.to[Array]
+                calculatedCrc32c = Crc32c.calculateCrc32c(fileBody)
+                syncStatus <- if (calculatedCrc32c == crc32c) IO.pure(SyncStatus.Live) else {
+                  for {
+                    cachedGeneration <- metadataCache.get.map(_.get(req.localObjectPath))
```
calbach: When does this cache get updated?

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -100,73 +103,92 @@ class ObjectService(
...
-                    else
-                      IO.pure(lastLockedBy)
-                  }
-                } yield MetadataResponse.EditMode(syncStatus, lastLock, generation, sl)
-            }
-          } yield result
-        }
+  def checkMetadata(req: GetMetadataRequest): Kleisli[IO, TraceId, MetadataResponse] = {
+    findStorageLink(
+      req.localObjectPath,
```
calbach: The new file resolution logic is working well now

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
Qi77Qi: it gets updated whenever we localize a file

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/package.scala
```diff
@@ -38,27 +38,24 @@ package object welder {
...
 
-  // base directory example: “workspaces/ws1”
-  def getLocalBaseDirectory(localPath: Path): Either[String, Path] = {
-    for {
-      prefix <- Either.catchNonFatal(localPath.getName(0)).leftMap(_ => s"no valid prefix found for $localPath")
-      workspaceName <- Either.catchNonFatal(localPath.getName(1)).leftMap(_ => s"no workspace name found for $localPath")
-    } yield localPath.subpath(0, 2)
+  def getPosssibleBaseDirectory(localPath: Path): List[Path] = {
+    ((localPath.getNameCount - 1).to(1, -1)).map(
+      x => localPath.subpath(0, x)
```
MatthewBemis: nit: can we be more descriptive than `x`?

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/Main.scala
```diff
@@ -17,27 +18,28 @@ import org.broadinstitute.dsp.workbench.welder.JsonCodec._
...
-      blockingEc <- Stream.resource[IO, ExecutionContext](ExecutionContexts.fixedThreadPool(255)) //TODO: use cached thread pool with some back pressure
-      storageLinksCache <- cachedResource[LocalDirectory, StorageLink](appConfig.pathToStorageLinksJson, blockingEc, storageLink => List(storageLink.localBaseDirectory -> storageLink, storageLink.localSafeModeBaseDirectory -> storageLink))
+      storageLinksCache <- cachedResource[Path, StorageLink](appConfig.pathToStorageLinksJson, blockingEc, storageLink => List(storageLink.localBaseDirectory.path -> storageLink, storageLink.localSafeModeBaseDirectory.path -> storageLink))
       metadataCache <- cachedResource[Path, GcsMetadata](appConfig.pathToGcsMetadataJson, blockingEc, metadata => List(metadata.localPath -> metadata))
-      _ <- Stream.eval(timer.sleep(2 seconds)) // sleep 2 seconds to wait for google application default credential becomes available
```
MatthewBemis: can you fill me in on what the solution for this was?

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -100,73 +105,107 @@ class ObjectService(
...
+      pair => Kleisli.liftF(IO.pure(MetadataResponse.SafeMode(pair.storageLink))),
```
MatthewBemis: i think people who aren't cats-savvy would benefit from a brief comment describing what `Kleisli` is and why it's used here.

in general i do worry about how FP-heavy welder is and how it could be challenging for somebody to pick it up and debug welder issues. this is probably a larger discussion but i think the more comments the better. this is a small repo, so i think it would be a good model-repo for showing off some of these FP approaches and comments would go a long way.

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -100,73 +105,107 @@ class ObjectService(
...
-            _ <- previousMeta match {
-              case Some(m) => delocalize(localAbsolutePath, gsPath, traceId, m.generation)
-              case None => delocalize(localAbsolutePath, gsPath, traceId, 0L)
-            }
-          } yield ()
-        }
+  def safeDelocalize(req: SafeDelocalize): Kleisli[IO, TraceId, Unit] = {
+    findStorageLink[Unit](
+      req.localObjectPath,
+      _ => Kleisli.liftF(IO.raiseError(SafeDelocalizeSafeModeFileError(s"${req.localObjectPath} can't be delocalized since it's in safe mode"))),
```
MatthewBemis: what is the benefit of refactoring `safeDelocalize` like this? i'm finding it significantly harder to understand

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/Main.scala
Qi77Qi: it was when Joel was working on deploying welder via leo and it can't find application default, yet it can find it if we restart welder service, so at the time, we thought a short delay before reading the application default will be good enough to make application default available on the box, but Joel ended doing what jupyter container is doing, which is overwriting startup script to a dummy command, and start welder service at some point later, which worked.

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
Qi77Qi: it's mostly to reduce duplicate code...`findStorageLink` is reused for `safeDelocalize`, `delete`, and `checkMeta`

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -100,98 +104,130 @@ class ObjectService(
...
-            fullBlobName <- IO.fromEither(getFullBlobName(req.localObjectPath, sl.cloudStorageDirectory.blobPath).leftMap(s => BadRequestException(s)))
-            gsPath = GsPath(sl.cloudStorageDirectory.bucketName, fullBlobName)
-            localAbsolutePath = config.workingDirectory.resolve(req.localObjectPath)
-            _ <- previousMeta match {
-              case Some(m) => delocalize(localAbsolutePath, gsPath, traceId, m.generation)
-              case None => delocalize(localAbsolutePath, gsPath, traceId, 0L)
-            }
+           previousMeta <- Kleisli.liftF(metadataCache.get.map(_.get(req.localObjectPath)))
+           _ <- previousMeta match {
+             case Some(m) => Kleisli.liftF[IO, TraceId, Unit](googleStorageService.removeObject(gsPath.bucketName, gsPath.blobName, Some(context.traceId)).void) //delete file from gcs with generation once https://github.com/broadinstitute/workbench-libs/pull/242 is approved
```
MatthewBemis: i think this comment is out of date now

PR #14, opened 2019-06-18
doc update

Qi77Qi:
minor clean up

```
~ curl -X POST --header 'Content-Type: application/json' --header 'Accept: application/json' localhost:8081/objects -d '{"action" : "localize",  "entries": [{ "sourceUri": "gs://qi-test/welder-test_nb.ipynb", "localDestinationPath": "/Users/jcanas/test" }] }'
{"errorMessage":"DecodingFailure(Path should be relative, List(DownField(localDestinationPath), DownArray, DownField(entries)))","errorCode":null}%
```


Qi77Qi:
oops..didn't realize this was approved...just added some minor improvement as well so that when user tries to use absolute path for `localObjectPath`, the request will fail

PR #15, opened 2019-06-19
[IA-1022] [IA-1063] add acquireLock endpoint

Qi77Qi:


core/src/main/scala/org/broadinstitute/dsp/workbench/welder/StorageLinksInterp.scala
```diff
@@ -0,0 +1,17 @@
+package org.broadinstitute.dsp.workbench.welder
+
+import cats.effect.IO
+
+class StorageLinksInterp(storageLinksCache: StorageLinksCache) extends StorageLinksAlg {
+  def findStorageLink[A](localPath: RelativePath): IO[CommonContext] = for {
+    storageLinks <- storageLinksCache.get
+    baseDirectories = getPosssibleBaseDirectory(localPath.asPath)
```
calbach: possible

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/StorageLinksInterp.scala
```diff
@@ -0,0 +1,17 @@
...
+    context = baseDirectories.collectFirst {
+      case x if (storageLinks.get(x).isDefined) =>
```
calbach: Checking: this map is indexed by both the edit mode base dir and the safe mode base dir?

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/StorageLinksInterp.scala
calbach: Is this ordered longest -> shortest?

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/GoogleStorageInterp.scala
```diff
@@ -0,0 +1,74 @@
...
+  def updateMetadata(gsPath: GsPath, traceId: TraceId, metadata: Map[String, String]): IO[Unit] = googleStorageService.setObjectMetadata(gsPath.bucketName, gsPath.blobName, metadata, Option(traceId)).compile.drain.handleErrorWith {
+    case e: com.google.cloud.storage.StorageException if(e.getCode == 403) =>
+      for {
+        _ <- logger.info(s"$traceId | Fail to update lock due to 403. Going to download the blob and re-upload")
+        bytes <- googleStorageService.getObjectBody(gsPath.bucketName, gsPath.blobName, Some(traceId)).compile.to[Array]
+        _ <- googleStorageService.storeObject(gsPath.bucketName, gsPath.blobName, bytes, "text/plain", metadata, None, Some(traceId)).compile.drain
+      } yield ()
+  }
+
+  def retrieveGcsMetadata(localPath: RelativePath, bucketName: GcsBucketName, blobName: GcsBlobName, traceId: TraceId): IO[Option[GcsMetadata]] = for {
```
MatthewBemis: this tripped me up a bit when i was reading the `acquireLock` function in `ObjectService`. this _is_ retrieving gcs metadata like it says but it's returning a transformed version of that metadata which isn't actually the gcs metadata anymore. what i was looking for in `acquireLock` was a condition for when another user's lock has expired but ended up having to come here to realize that the lock metadata is thrown out if it's expired.

could you rename this function to accurately reflect what it's really doing?

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/StorageLinksInterp.scala
Qi77Qi: yes...see https://github.com/DataBiosphere/welder/blob/master/server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/StorageLinksService.scala#L39

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/StorageLinksInterp.scala
Qi77Qi: yes..see https://github.com/DataBiosphere/welder/blob/master/core/src/main/scala/org/broadinstitute/dsp/workbench/welder/package.scala#L50, the code is a bit confusing, but this might help
```
scala> 10.to(1, -1).toList
res1: List[Int] = List(10, 9, 8, 7, 6, 5, 4, 3, 2, 1)
```

`-1` is step.

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/StorageLinksAlg.scala
```diff
@@ -0,0 +1,14 @@
+package org.broadinstitute.dsp.workbench.welder
+
+import java.nio.file.Path
+
+import cats.effect.IO
+
+trait StorageLinksAlg {
```
MatthewBemis: seen this a few times but is `alg` short for `algebra`?

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/GoogleStorageInterp.scala
```diff
@@ -0,0 +1,85 @@
...
+import org.broadinstitute.dsde.workbench.google2
+import org.broadinstitute.dsde.workbench.google2.{Crc32, GoogleStorageService, RemoveObjectResult}
+import org.broadinstitute.dsde.workbench.model.{TraceId, WorkbenchEmail}
+import org.broadinstitute.dsp.workbench.welder.GoogleStorageAlg._
+import org.broadinstitute.dsp.workbench.welder.SourceUri.GsPath
+import scala.collection.JavaConverters._
+
+class GoogleStorageInterp(config: GoogleStorageAlgConfig , googleStorageService: GoogleStorageService[IO])
+                         (implicit logger: Logger[IO], timer: Timer[IO], linerBacker: Linebacker[IO], cs: ContextShift[IO]) extends GoogleStorageAlg {
+  def updateMetadata(gsPath: GsPath, traceId: TraceId, metadata: Map[String, String]): IO[Unit] = googleStorageService.setObjectMetadata(gsPath.bucketName, gsPath.blobName, metadata, Option(traceId)).compile.drain.handleErrorWith {
```
MatthewBemis: nit: can you bump the first line of this function onto the next line?

the same comment will apply to a few of the others. it helps with readability IMO and is currently inconsistent with other functions in this class

Qi77Qi: sure yea..I'll run scalafmt, it should fix this

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/StorageLinksAlg.scala
Qi77Qi: yea...it is...let me know if you have a better name.

PR #19, opened 2019-06-25
IA-1092 bug fix

Qi77Qi:
- Fix a bug where lastLockedBy field wasn't checked properly in acquireLock endpoint
- Add unit tests for each case


server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -106,12 +106,17 @@ class ObjectService(
...
+              if(lockedBy == config.ownerEmail)
+                googleStorageAlg.updateMetadata(gsPath, traceId, metadata).as(AcquireLockResponse.Success)
+              else
+                IO.pure(AcquireLockResponse.LockedByOther)
+            case None =>
+              googleStorageAlg.updateMetadata(gsPath, traceId, metadata).as(AcquireLockResponse.Success)
+          }
         case None =>
-          googleStorageAlg.updateMetadata(gsPath, traceId, metadata).as(AcquireLockResponse.Success)
+          IO.raiseError(NotFoundException(s"${gsPath} not found in Google Storage")) //TODO: is this right?
```
MatthewBemis: i don't think it's necessarily wrong. you shouldn't be able to lock objects that aren't covered by storage links and if it's covered by a storage link it should be getting delocalized to gcs. have we seen this in testing? 

an alternative could be to delocalize it with the lock metadata if it's covered by a storage link.

calbach: Is there a case where we are actually trying to acquire a lock for a non-existent file? Wouldn't this indicate a bug?

Qi77Qi: file here will always be covered by storagelinks cuz `findStorageLink` before this step.

> Is there a case where we are actually trying to acquire a lock for a non-existent file? Wouldn't this indicate a bug?
What about the first time a file gets delocalized, do we need to have a lock to do that? (maybe current implementation is correct since in safeDelocalize we don't check lock; hence there's no need to acquireLock before safeDelocalize)

project/Settings.scala
```diff
@@ -104,6 +104,7 @@ object Settings {
   lazy val serverDockerSettings = commonDockerSettings ++ List(
     mainClass in Compile := Some("org.broadinstitute.dsp.workbench.welder.server.Main"),
     packageName in Docker := "broad-dsp-gcr-public/welder-server",
+//    daemonUser in Docker := "welder-user",
```
calbach: Did you mean to include this?

Qi77Qi: No. It’s not working yet. I’ll try it tomorrow

PR #20, opened 2019-06-25
Hash email with bucket name before storing in object metadata

MatthewBemis:


server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -98,18 +99,24 @@ class ObjectService(
...
       )
-      gsPath = getGsPath(req.localObjectPath, context)
       meta <- googleStorageAlg.retrieveAdaptedGcsMetadata(req.localObjectPath, gsPath, traceId)
       res <- meta match {
         case Some(m) =>
-          if (m.lastLockedBy.isDefined && m.lastLockedBy.get == config.ownerEmail)
-            googleStorageAlg.updateMetadata(gsPath, traceId, metadata).as(AcquireLockResponse.Success)
-          else
-            IO.pure(AcquireLockResponse.LockedByOther)
+          m.lastLockedBy match {
```
MatthewBemis: this is lifted from #19

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/package.scala
```diff
@@ -75,6 +77,10 @@ package object welder {
     GsPath(basePathAndStorageLink.storageLink.cloudStorageDirectory.bucketName, fullBlobName)
   }
 
+  def hashMetadata(metadata: String): HashedMetadata = {
+    HashedMetadata(String.format("%032x", new BigInteger(1, MessageDigest.getInstance("SHA-256").digest(metadata.getBytes("UTF-8")))))
```
Qi77Qi: I would just call this `hasString` since it has nothing to do with `metadata` in particular.
Also `String.format` may throw exception. So we should wrap it in `IO`

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/model.scala
```diff
@@ -68,10 +67,12 @@ final case class StorageLink(
     pattern: String
 )
 
+final case class HashedMetadata(asString: String) extends AnyVal
```
Qi77Qi: shall we call this `hashedLockedBy` instead? cuz it's not hashing all metadata fields

MatthewBemis: 👍 I wanted it to be more generic because there were talks of having a lastEditedBy field which would also need to be hashed, but I don't think that's actually happening so this ended up being premature

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/GoogleStorageInterp.scala
```diff
@@ -72,7 +72,7 @@ class GoogleStorageInterp(config: GoogleStorageAlgConfig, googleStorageService:
 
   private def adaptMetadata(crc32c: Crc32, userDefinedMetadata: Map[String, String], generation: Long): IO[AdaptedGcsMetadata] =
     for {
-      lastLockedBy <- IO.pure(userDefinedMetadata.get(LAST_LOCKED_BY).map(WorkbenchEmail))
+      lastLockedBy <- IO.pure(userDefinedMetadata.get(LAST_LOCKED_BY).map(HashedMetadata))
```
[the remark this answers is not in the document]
MatthewBemis: we shouldn't need to decode back. it's being hashed in gcs metadata which should then cause it to be pulled down into the metadata cache, so it's hashed both places, yes

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -98,17 +99,18 @@ class ObjectService(
     for {
       context <- storageLinksAlg.findStorageLink(req.localObjectPath)
       current <- timer.clock.realTime(TimeUnit.MILLISECONDS)
+      gsPath = getGsPath(req.localObjectPath, context)
+      hashedLockedByCurrentUser <- hashString(lockedByString(gsPath.bucketName, config.ownerEmail))
       metadata = Map(
-        GoogleStorageAlg.LAST_LOCKED_BY -> config.ownerEmail.value,
+        GoogleStorageAlg.LAST_LOCKED_BY -> hashedLockedByCurrentUser.asString,
```
Qi77Qi: there're a few different places where `config.ownerEmail` is used. I'm thinking it's better to change `ownerEmail` to `HashedLockedBy` in config, and do the hashing in https://github.com/DataBiosphere/welder/blob/master/server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/Config.scala#L21

MatthewBemis: i agree it would be nice to push that up to config.scala but the hashing is a combo of the gcs bucket name and the user's email. since we won't know which bucket they're syncing to until the lock request comes in (and there could be multiple buckets that welder is juggling), this wouldn't be possible

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -98,17 +99,18 @@ class ObjectService(
     for {
       context <- storageLinksAlg.findStorageLink(req.localObjectPath)
       current <- timer.clock.realTime(TimeUnit.MILLISECONDS)
+      gsPath = getGsPath(req.localObjectPath, context)
+      hashedLockedByCurrentUser <- hashString(lockedByString(gsPath.bucketName, config.ownerEmail))
```
calbach: Does this include the gs:// or not? I don't care either way, but the details should probably go into the README or the API yaml docs so clients have something to reference to unpack this value.

MatthewBemis: it doesn't include the `gs://` prefix. i'll document.

calbach: Thanks. Also I misspoke by saying "unpack"; clients will need to generate the hash in the same way as Welder (but by design, the hash is not "unpackable").

PR #21, opened 2019-06-26
IA-1094 bug fixes

Qi77Qi:
* Update lock everytime we interacts with google
* update crc32c and generation in both localizing and delocalizing
* push old lock when delocalizing instead of wiping out old lock
* ignore welder-user related changes for now...Rob will have a separate PR
* Added an infinite Stream that'll clean up expired locks in metadata cache every 7 minutes along side the web server. It currently will log the entire metadata cache...
```
09:05:12.728 [ioapp-compute-2] INFO org.broadinstitute.dsp.workbench.welder.server.Main - updated metadata cache Map(qi-test2.ipnyb -> AdaptedGcsMetadataCache(qi-test2.ipnyb,None,None), qi/Users/jcanas/test -> AdaptedGcsMetadataCache(qi/Users/jcanas/test,None,Some(LocalFileStateInGCS(Crc32(SnPQRQ==),1560777047826706))))
09:05:42.731 [ioapp-compute-0] INFO org.broadinstitute.dsp.workbench.welder.server.Main - updated metadata cache Map(qi-test2.ipnyb -> AdaptedGcsMetadataCache(qi-test2.ipnyb,None,None), qi/Users/jcanas/test -> AdaptedGcsMetadataCache(qi/Users/jcanas/test,None,Some(LocalFileStateInGCS(Crc32(SnPQRQ==),1560777047826706))))
```


project/Settings.scala
```diff
@@ -98,13 +99,16 @@ object Settings {
     dockerUpdateLatest := true,
     dockerCommands ++= List(
       ExecCmd("CMD", "export", "CLASSPATH=lib/*jar")
+//      ExecCmd("CMD", "unmask", "umask g+w") //TODO: doesn't seem to do anything yet
```
rtitle: Based on @calbach's comment in https://broadworkbench.atlassian.net/browse/IA-1086 it seems like the umask needs to be in the `ENTRYPOINT`. I was wondering if another option would be to append it to `~/.bashrc`?

project/Settings.scala
```diff
@@ -98,13 +99,16 @@ object Settings {
...
     )
   )
 
   lazy val serverDockerSettings = commonDockerSettings ++ List(
     mainClass in Compile := Some("org.broadinstitute.dsp.workbench.welder.server.Main"),
     packageName in Docker := "broad-dsp-gcr-public/welder-server",
-//    daemonUser in Docker := "welder-user",
+    daemonUser in Docker := "welder-user",
+    daemonGroup in Docker := "user",
```
rtitle: `users` is the group i think

project/Settings.scala
```diff
@@ -98,13 +99,16 @@ object Settings {
...
+    daemonGroupGid in Docker := Some("25000"),
```
rtitle: Should the gid be consistent with the Jupyter image? It's 100 there I think.

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -106,15 +106,18 @@ class ObjectService(
       meta <- googleStorageAlg.retrieveAdaptedGcsMetadata(req.localObjectPath, gsPath, traceId)
       res <- meta match {
         case Some(m) =>
-          m.lastLockedBy match {
-            case Some(lockedBy) =>
-              if(lockedBy == config.ownerEmail)
+          for {
```
rtitle: Not sure but this code seems unrelated to IA-1086?

project/Settings.scala
Qi77Qi: yea, I was wondering if I should use gid in jupyter, but I didn't know about it...So just used a random one as stub

project/Settings.scala
calbach: Things I also tried for the umask which didn't work for me:

- Append to `~/.bashrc`
- Append to `/etc/profile`
- In the Dockerfile: `umask 0000`

server/src/main/resources/api-docs.yaml
```diff
@@ -183,12 +183,18 @@ definitions:
         type: string
       errorCode:
         type: Int
-        description: File generation mismatch(0), no storage link found for the requested file (1), trying to delocalize safe mode file (2), trying to delete safe mode file (3)
+        description: |
+          0: File generation mismatch
+          1: no storage link found for the requested file
+          2: trying to delocalize safe mode file
+          3: trying to delete safe mode file
+          4: current user doesn't hold a lock the file
```
calbach: nit: "for the file"

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -61,6 +63,7 @@ class ObjectService(
       } yield res
   }
 
+  //TODO: do we need to check storagelink before localizing a file?
```
calbach: Ideally /localize is a general purpose endpoint that can be used independently.

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -144,21 +147,31 @@ class ObjectService(
...
-                      case Some(cachedGen) => if (cachedGen.generation == generation) IO.pure(SyncStatus.LocalChanged) else IO.pure(SyncStatus.RemoteChanged)
-                      case None => IO.pure(SyncStatus.Desynchronized)
+                    metaOpt <- metadataCache.get.map(_.get(req.localObjectPath))
+                    status <- metaOpt match {
+                      case Some(meta) =>
+                        meta.localFileStateInGCS match { //TODO: shall we check the file has been localized before calculating crc32c
+                          case Some(previousFileState) =>
+                            if (previousFileState.generation == generation)
+                              IO.pure(SyncStatus.LocalChanged) <* logger.ctxWarn[PostContext, String]("local file has changed, but it hasn't been delocalized yet").run(loggingContext)
+                            else IO.pure(SyncStatus.RemoteChanged) <* logger.ctxWarn[PostContext, String]("remote has changed").run(loggingContext)
```
calbach: Including the generations in this log line would be useful

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -144,21 +147,31 @@ class ObjectService(
...
+                          case None =>
+                            IO.pure(SyncStatus.Desynchronized) <* logger.ctxError[PostContext, String]("We don't find local generation for a localized file. This shouldn't happen").run(loggingContext)
+                        }
+                      case None =>
+                        IO.pure(SyncStatus.Desynchronized) <* logger.ctxError[PostContext, String]("We don't find local cache for a localized file. This shouldn't happen").run(loggingContext) //TODO: this shouldn't be possible because we should have the file in cache if it has been localized
```
calbach: There are many ways this could happen, e.g. transient failures, direct file system manipulation on the VM (`cp` from terminal, `!touch` from notebook, etc). Ideally Welder should handle this scenario gracefully if it wants to be more robust. This goes back to the metadata caching discussion - I would change the current behavior as follows:

- metadata cache always tracks the latest known GCS metadata
- /checkMetadata doesn't check the cache, it always reads from GCS and then updates the cache accordingly
- Everything else can reference the cache for reading metadata values (optionally forcing a GET if we've never cached the value yet)
- Don't touch the cache in /localize

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -193,13 +215,19 @@ class ObjectService(
...
+        val res = for {
+          previousMeta <- metadataCache.get.map(_.get(req.localObjectPath))
+          now <- timer.clock.realTime(TimeUnit.MILLISECONDS)
+          generation <- previousMeta match {
+            case Some(meta) =>
+              checkLock(meta.lock, now, gsPath.bucketName).as(meta.localFileStateInGCS.map(_.generation).getOrElse(0L)) //check if user owns the lock before deleting
+            case None =>
+              IO.raiseError(InvalidLock(s"Local GCS metadata for ${req.localObjectPath} not found"))
           }
+          _ <- googleStorageAlg.removeObject(gsPath, traceId, Some(generation)).compile.drain.void
```
calbach: I don't follow the Scala - if this fails do we throw an exception here?

project/Settings.scala
```diff
@@ -98,13 +99,16 @@ object Settings {
...
       ExecCmd("CMD", "export", "CLASSPATH=lib/*jar")
+//      ExecCmd("CMD", "unmask", "umask g+w") //TODO: doesn't seem to do anything yet
     )
   )
 
   lazy val serverDockerSettings = commonDockerSettings ++ List(
     mainClass in Compile := Some("org.broadinstitute.dsp.workbench.welder.server.Main"),
     packageName in Docker := "broad-dsp-gcr-public/welder-server",
-//    daemonUser in Docker := "welder-user",
+    daemonUser in Docker := "welder-user",
```
calbach: Please revert so you do not accidentally merge this

rtitle: My PR (https://github.com/DataBiosphere/welder/pull/23) updates this, can probably revert all `Settings.scala` changes here.

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
Qi77Qi: yes

project/Settings.scala
Qi77Qi: This will be taken care of after rebasing

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
Qi77Qi: If we don't update cache's generation when localizing, how do we know the generation before user modifies the file? That generation is used for delocalizing and deleting

calbach: Here is what I would propose:

```
safeDelocalize:
  meta = metaCache.get(path, fetchIfMissing=true)
```

It wouldn't hurt to cache on localization, but I wouldn't want to depend on that. I think it's a risky approach to try and control all mechanisms by which a file could get onto the VM.

If we did that, we'd also need to slightly change semantics (as we discussed previously: https://github.com/DataBiosphere/leonardo/blob/develop/docker/jupyter/custom/jupyter_delocalize.py#L217) - Jupyter should not update the file until after Welder. Then `/safeDelocalize` could compare the current crc32cs before deciding whether to update.

calbach: @Qi77Qi convinced me in Slack. There are downsides to both approaches but the current way seems likely to behave better during standard usage.

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/ContextLogger.scala
```diff
@@ -0,0 +1,53 @@
...
+import io.chrisdavenport.log4cats.Logger
+import io.circe.Encoder
+import io.circe.syntax._
+
+/**
+  * Provides syntax extension for `Logger[F]`
+  *
+  * `ctxXXX` logs context `A` together with a given message
+  */
+final case class ContextLogger[F[_]](log: Logger[F]) extends AnyVal {
```
Qi77Qi: This class makes it easier to write log lines that share common context

server/src/main/resources/logback.xml
```diff
@@ -11,12 +11,33 @@
     <encoder>
       <pattern>%date %level [%thread] %logger{10} [%file:%line] %msg%n</pattern>
     </encoder>
+    <rollingPolicy class="ch.qos.logback.core.rolling.TimeBasedRollingPolicy">
+      <!-- daily rollover -->
+      <fileNamePattern>logs\logFile.%d{yyyy-MM-dd}.%i.html</fileNamePattern>
```
rtitle: `.html`? Curious, why html?

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/Main.scala
```diff
@@ -21,58 +16,38 @@ import scala.concurrent.ExecutionContext
...
+      serverStream = BlazeServerBuilder[IO].bindHttp(appConfig.serverPort, "0.0.0.0").withHttpApp(WelderApp(objectService, storageLinksService).service).serve
+      cleanUpCache = BackgroundTask.cleanUpLock(metadataCache, appConfig.cleanUpLockFrequency)
+      _ <- Stream(cleanUpCache, serverStream.drain).covary[IO].parJoin(2)
     } yield ()
 
     app
       .handleErrorWith { error =>
         Stream.eval(Logger[IO].error(error)("Failed to start server")) >> Stream.raiseError[IO](error)
       }
-      .evalMap(_ => IO.never)
```
rtitle: Why don't we need this anymore? Is it because we now have the infinite `BackgroundTask` stream?

server/src/main/resources/logback.xml
Qi77Qi: ah...good catch! This was copy pasted from stackoverflow for rotating logs...

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/Main.scala
Qi77Qi: yep...the background task will prevent the app from exiting, so we don't really need this...The background task won't actually work with this because background process is an infinite Stream, and this will prevent the first emitted IO to not return.

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/BackgroundTask.scala
```diff
@@ -0,0 +1,30 @@
...
+    val task = (for {
+      now <- timer.clock.monotonic(TimeUnit.MILLISECONDS)
+      updatedMap <- metadataCache.modify { mp =>
+        val newMap = mp.map { kv =>
+          val newLock = kv._2.lock.filter(l => l.lockExpiresAt.toEpochMilli < now)
+          (kv._1 -> kv._2.copy(lock = newLock))
+        }
+        (newMap, newMap)
+      }
+      _ <- logger.info(s"updated metadata cache ${updatedMap.take(100)}")
```
rtitle: Wondering if it's ok to log the cache... we don't want to inadvertently fill up the disk (that the user is paying for) with these log messages. Maybe we could log something like `"updated metadata cache with n entries"` at INFO level and log the actual cache at DEBUG level.

Qi77Qi: Good idea..updated

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/BackgroundTask.scala
```diff
@@ -0,0 +1,31 @@
...
+      now <- timer.clock.monotonic(TimeUnit.MILLISECONDS)
+      updatedMap <- metadataCache.modify { mp =>
+        val newMap = mp.map { kv =>
+          val newLock = kv._2.lock.filter(l => l.lockExpiresAt.toEpochMilli < now)
+          (kv._1 -> kv._2.copy(lock = newLock))
+        }
+        (newMap, newMap)
+      }
+      _ <- logger.info(s"updated metadata cache size ${updatedMap.size}")
+      _ <- logger.debug(s"updated metadata cache ${updatedMap.size}")
```
rtitle: Debug message still has `.size`

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/package.scala
```diff
@@ -82,14 +84,35 @@ package object welder {
...
+  type StorageLinksCache = Ref[IO, Map[RelativePath, StorageLink]]
+  type MetadataCache = Ref[IO, Map[RelativePath, AdaptedGcsMetadataCache]]
 
   val gcpObjectType = "text/plain"
 
+  def cachedResource[A, B: Decoder: Encoder](path: Path, blockingEc: ExecutionContext, toTuple: B => List[(A, B)])(
+    implicit logger: Logger[IO], cs: ContextShift[IO]
+  ): Stream[IO, Ref[IO, Map[A, B]]] =
+    for {
+      cached <- readJsonFileToA[IO, List[B]](path).map(ls => ls.flatMap(b => toTuple(b)).toMap).handleErrorWith { error =>
```
rtitle: Is there a risk of losing data here? If there are key collisions, `listOfTuples.toMap` will throw away keys. Do we want some kind of `groupBy` instead?

Ex:
```
// imagine my dataset is 6 ints
scala> val dataset = List(1,2,3,4,5,6)
dataset: List[Int] = List(1, 2, 3, 4, 5, 6)

// I want to group them by odd or even
scala> def toTuple(i: Int): List[(Int, Int)] = List(i % 2 -> i)
toTuple: (i: Int)List[(Int, Int)]

// I've lost most of my data set
scala> dataset.flatMap(toTuple).toMap
res6: scala.collection.immutable.Map[Int,Int] = Map(1 -> 5, 0 -> 6)
```

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/package.scala
```diff
@@ -82,14 +84,35 @@ package object welder {
...
+        Stream.eval(logger.info(s"$path not found")) >> Stream.emit(Map.empty[A, B]).covary[IO]
+      }
+      ref <- Stream.resource(
+        Resource.make(Ref.of[IO, Map[A, B]](cached))(
+          ref =>
```
rtitle: Having trouble sorting out the streams -- when does this release function run? Is it after each usage of the cached resource, after every API request, or at the end of the app (e.g. when welder shuts down)? I want to make sure we're reliably flushing caches to disk and not losing any data if for example the VM gets paused.

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/BackgroundTask.scala
Qi77Qi: updated

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/package.scala
Qi77Qi: Every time the app starts up, we load the file into memory. Release should be guaranteed to run when app shuts down.

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/package.scala
Qi77Qi: I think for one particular local file path, we should only have 1 storagelink, and 1 metadata cache. New data should overwrite old record.

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/package.scala
rtitle: Does the app shut down cleanly though when the cluster is stopped? Might be worth a test.

Qi77Qi: Sure...https://broadworkbench.atlassian.net/browse/IA-1107

PR #23, opened 2019-06-28
[IA-1086] Reconcile shared Welder volume ownership/permissions

rtitle:
Friends with: https://github.com/DataBiosphere/leonardo/pull/977
JIRA: https://broadworkbench.atlassian.net/browse/IA-1086

This PR essentially does 2 things:
1. Replaces the welder docker `ENTRYPOINT` with a new `entrypoint.sh` script which sets a umask prior to starting the welder server.
2. Configures welder to run as `welder-user:users`. This user/group is replicated in the Jupyter container, so welder-created files display as `welder-user`.

The result of this is that files/directories in the shared volume are always R/W to both containers. I verified this manually in both directions.

### DIFF
--- a/project/Settings.scala
+++ b/project/Settings.scala
@@ -2,11 +2,12 @@ import java.time.ZoneId
 
 import com.typesafe.sbt.SbtNativePackager.autoImport._
 import com.typesafe.sbt.packager.docker.DockerPlugin.autoImport._
-import com.typesafe.sbt.packager.docker.ExecCmd
+import com.typesafe.sbt.packager.docker.{Cmd, ExecCmd}
 import org.scalafmt.sbt.ScalafmtPlugin.autoImport.scalafmtOnCompile
 import sbt.Keys.{scalacOptions, _}
 import sbt._
 import com.typesafe.sbt.GitPlugin.autoImport._
+import com.typesafe.sbt.packager.linux.LinuxPlugin.autoImport._
 import sbtbuildinfo.BuildInfoPlugin.autoImport._
 import com.gilt.sbt.newrelic.NewRelic.autoImport._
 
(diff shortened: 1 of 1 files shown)


project/Settings.scala
```diff
@@ -90,22 +91,33 @@ object Settings {
...
+
   lazy val commonDockerSettings = List(
     maintainer := "workbench-interactive-analysis@broadinstitute.org",
     dockerBaseImage := "oracle/graalvm-ce:1.0.0-rc16",
     dockerRepository := Some("us.gcr.io"),
     dockerExposedPorts := List(8080),
     dockerUpdateLatest := true,
+    // See https://www.scala-sbt.org/sbt-native-packager/formats/docker.html#add-commands
     dockerCommands ++= List(
-      ExecCmd("CMD", "export", "CLASSPATH=lib/*jar")
```
rtitle: I removed this `CMD` because it seemed like the classpath was already set, and it doesn't seem like the correct usage of docker `CMD` anyway

project/Settings.scala
```diff
@@ -90,22 +91,33 @@ object Settings {
...
+      // Change the default umask for welder to support R/W access to the shared volume
+      Cmd("USER", "root"),
+      ExecCmd("RUN", "/bin/bash", "-c", s"echo '#!/bin/bash\\numask 002; /opt/docker/bin/server' > $entrypoint"),
+      Cmd("RUN", s"chown -R welder-user:users /opt/docker && chmod u+x,g+x $entrypoint"),
```
Qi77Qi: Does `g` need permission to execute welder?

project/Settings.scala
Qi77Qi: hmmm weird..I thought I ran into jar can't find without this.

project/Settings.scala
rtitle: No but I tried to be consistent — the sbt plugin also grants u+x,g+x to the /opt/docker/bin/server executable

PR #25, opened 2019-07-03
add flush api

Qi77Qi:
when cluster gets shut down, jvm shutdown hook doesn't seem to get executed. Add an API for explicitly flushing cache to disk

### DIFF
--- a/server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/Main.scala
+++ b/server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/Main.scala
@@ -1,8 +1,6 @@
 package org.broadinstitute.dsp.workbench.welder
 package server
 
-import java.nio.file.Paths
-
 import cats.effect.{ExitCode, IO, IOApp}
 import cats.implicits._
 import fs2.Stream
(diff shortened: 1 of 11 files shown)


server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/Main.scala
```diff
@@ -32,12 +32,8 @@ object Main extends IOApp {
         blockingEc,
         metadata => List(metadata.localPath -> metadata)
       )
-      googleStorageService <- Stream.resource(GoogleStorageService.fromApplicationDefault())
```
rtitle: ^ `implicit0(it: Linebacker[IO]) <-` can probably be removed here right, since it's just needed for `GoogleStorageService`?

Qi77Qi:
ah good idea..I'll add flushing periodically

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/Main.scala
Qi77Qi: updated

rtitle:
Latest change looks good :shipit:

PR #27, opened 2019-07-03
Jc storagelinks init

jdcanas:


server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/StorageLinksService.scala
```diff
@@ -9,8 +9,13 @@ import org.http4s.HttpRoutes
...
 import org.http4s.circe.CirceEntityEncoder._
 import org.http4s.dsl.Http4sDsl
+import java.nio.file.Path
 
-class StorageLinksService(storageLinks: StorageLinksCache) extends Http4sDsl[IO] {
+import io.chrisdavenport.log4cats.Logger
+import io.chrisdavenport.log4cats.slf4j.Slf4jLogger
+
+class StorageLinksService(storageLinks: StorageLinksCache, workingDirectory: Path) extends Http4sDsl[IO] {
+  implicit val logger: Logger[IO] = Slf4jLogger.getLogger[IO]
```
Qi77Qi: pass logger in implicitly like `(implicit cs: Logger[IO])`

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/StorageLinksService.scala
```diff
@@ -35,10 +40,27 @@ class StorageLinksService(storageLinks: StorageLinksCache) extends Http4sDsl[IO]
   //note: first param in the modify is the thing to do, second param is the value to return
   def createStorageLink(storageLink: StorageLink): IO[StorageLink] =
     storageLinks.modify { links =>
+
       val toAdd = List(storageLink.localBaseDirectory.path -> storageLink, storageLink.localSafeModeBaseDirectory.path -> storageLink).toMap
+
+      initializeDirectories(storageLink)
```
Qi77Qi: you don't want to stick the side effect inside modify storagelinks. Instead chain the effect after modify
something like
```
for {
_ <- storageLinks.modify { links =>xxx}
_ <- initializeDirectories(storageLink)
} yield xxx

File: server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/StorageLinksService.scala [hunk 3]
```diff
@@ -35,10 +40,27 @@ class StorageLinksService(storageLinks: StorageLinksCache) extends Http4sDsl[IO]
...
+  def initializeDirectories(storageLink: StorageLink): Boolean = {
+    val localSafeAbsolutePath = workingDirectory.resolve(storageLink.localSafeModeBaseDirectory.path.asPath)
+    val localEditAbsolutePath = workingDirectory.resolve(storageLink.localBaseDirectory.path.asPath)
+
+    val dirsToCreate: List[java.nio.file.Path] = List[java.nio.file.Path](localSafeAbsolutePath, localEditAbsolutePath)
+
+    dirsToCreate
+      .map(path => new java.io.File(path.toUri))
+      .map(dir => dir.exists() || dir.mkdir()) //mkdir returns true if the directory is created. if it exists it returns false.
+      .foldRight(true)(_ && _)
```
Qi77Qi: making directory is a side effect. wrap the action where actually creating directory in an `IO`, and return `IO[Unit]` (doesn't look like you care whether the directory is created successfully or not at call site)

PR #33, opened 2019-07-08
IA-1155 IA-1150 

Qi77Qi:
* IA-1155 return 404 when localizing a non existent file #33
* Fix a bug where we can't delocalize a new file because we require a lock, but lock can't be obtained when a file doesn't exist in gcs

### DIFF
--- a/server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
+++ b/server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
@@ -73,8 +73,8 @@ class ObjectService(
           case DataUri(data) => Stream.emits(data).through(io.file.writeAll[IO](localAbsolutePath, blockingEc))
           case gsPath: GsPath =>
             for {
-              meta <- googleStorageAlg.gcsToLocalFile(localAbsolutePath, gsPath, traceId)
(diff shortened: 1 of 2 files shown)


server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -213,7 +213,12 @@ class ObjectService(
           lockToPush <- previousMeta match {
             case Some(meta) =>
               checkLock(meta.lock, now, gsPath.bucketName)
-            case None => IO.raiseError(InvalidLock("Lock not found in local cache. You should acquireLock more often"))
+            case None =>
```
calbach: I don't know best practices in Scala, but yielding `(lockMeta, generation)` from here would be clearer than having the logic in two places, then you just return `lockMeta, 0`

Qi77Qi: it wouldn't make this cleaner since `checkLock(meta.lock, now, gsPath.bucketName)` is an `IO[Map[String, String]]` while generation is a pure value. So in order to yield `(lockMeta, generation)`, you need to lift `generation` into an `IO` first, which is unnecessary.

PR #34, opened 2019-07-09
IA-1104 move application.log to working directory

Qi77Qi:
Tested locally in docker image

### DIFF
--- a/server/src/main/resources/logback.xml
+++ b/server/src/main/resources/logback.xml
@@ -1,6 +1,8 @@
 <configuration>
+  <property name="WORKING_DIRECTORY" value="/work" />
+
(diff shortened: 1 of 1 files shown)


server/src/main/resources/logback.xml
```diff
@@ -1,6 +1,8 @@
 <configuration>
+  <property name="WORKING_DIRECTORY" value="/work" />
+
   <appender name="FILE" class="ch.qos.logback.core.FileAppender">
-    <file>application.log</file>
+    <file>${WORKING_DIRECTORY}/welder.log</file>
```
rtitle: Think the `fileNamePattern` below needs updating too?

Qi77Qi: thanks!..updated

PR #37, opened 2019-07-16
IA-1153 make localize always overwrite existing file

Qi77Qi:


core/src/main/scala/org/broadinstitute/dsp/workbench/welder/GoogleStorageInterp.scala
```diff
@@ -55,7 +57,7 @@ class GoogleStorageInterp(config: GoogleStorageAlgConfig, googleStorageService:
       _ <- (Stream
         .emits(blob.getContent())
         .covary[IO]
-        .through(io.file.writeAll[IO](localAbsolutePath, linerBacker.blockingContext))) ++ Stream.eval(IO.unit)
+        .through(io.file.writeAll[IO](localAbsolutePath, linerBacker.blockingContext, List(StandardOpenOption.TRUNCATE_EXISTING)))) ++ Stream.eval(IO.unit)
```
calbach: Will this still create the file if it doesn't already exist, or do you need to include CREATE as well?

rtitle: Just wondering if this should be:
```
List(StandardOpenOption.CREATE, StandardOpenOption.TRUNCATE_EXISTING)
```
With no parameters it defaults to `List(StandardOpenOption.CREATE)`. See: https://oss.sonatype.org/service/local/repositories/releases/archive/co/fs2/fs2-io_2.12/1.0.1/fs2-io_2.12-1.0.1-javadoc.jar/!/fs2/io/file/index.html

StandardOpenOption reference: https://docs.oracle.com/javase/7/docs/api/java/nio/file/StandardOpenOption.html

Qi77Qi: I’ll check test tomorrow. If ‘Create’ is needed, the other test will fail

PR #38, opened 2019-07-19
IA-1194 don't delocalize when there's no local change

Qi77Qi:


Qi77Qi:
@rtitle , yea the refactoring is mostly for caching remote crc32 locally so that we can compare current file crc32c against the cached crc32c to see if we have local changes.

PR #41, opened 2019-07-23
locks expire after 5mins now

jdcanas:


Qi77Qi:
Shall we just change this to 3 min?

calbach:
I like 3min if there are no objections.

server/src/main/resources/reference.conf
```diff
@@ -3,7 +3,7 @@ path-to-storage-links-json = "/work/.welder/storage_links.json"
 path-to-gcs-metadata-json = "/work/.welder/gcs_metadata.json"
 object-service {
   working-directory = "/work"
-  lock-expiration = 20 minutes
+  lock-expiration = 3 minutes
   owner-email = ${OWNER_EMAIL}
 }
 clean-up-lock-interval = 7 minutes
```
calbach: Doesn't this need to be shorter than the lock-expiration?

Qi77Qi: No, this is more of nice to have optimization. If it’s shorter than lock expiration, then we’ll be trying to clean up lock more often we need to

PR #42, opened 2019-08-07
miscellaneous changes

Qi77Qi:
1. upgrade scalafmt to 2.0.3
2. run scalafmt
3. upgrade http4s to 0.20.9
4. downgrade sbt to 1.2.8

### DIFF
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=1.3.0-RC2
+sbt.version=1.2.8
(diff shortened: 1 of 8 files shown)


project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.3.0-RC2
+sbt.version=1.2.8
```
Qi77Qi: downgrade to 1.2.8 because 1.3.0-RC2 hangs repl when running server locally

PR #43, opened 2019-08-09
IA-1289 use traceId from stackdriver if available

Qi77Qi:
This PR has two changes

- https://broadworkbench.atlassian.net/browse/IA-1289. For people who're not familiar with Semaphore, check out [this](https://typelevel.org/cats-effect/concurrency/semaphore.html). By using semaphore, we can prevent `checkMeta` and `safeDelocalize` happen at the same time. I'm adding semaphore to `/lock` api as well since all these 3 APIs touch the local cache.

- Use `X-Cloud-Trace-Context` as traceId when it's available. This is dependent on https://github.com/broadinstitute/workbench-libs/pull/268 which is blocked due to lack of review, hence using a SNAP version for now, and we can update the dependency when the wb-lib PR is approved

AOU is using stackdriver tracing, which injects `X-Cloud-Trace-Context`. This PR will utilize this uuid when it exists
```
2019-08-08 20:00:43,295 INFO [ioapp-compute-0] o.h.s.m.Logger [Logger.scala:29] HTTP/1.1 POST /objects/lock Headers(Host: 127.0.0.1:8080, Origin: https://leonardo.dsde-dev.broadinstitute.org, User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/75.0.3770.142 Safari/537.36, Accept: */*, Referer: https://leonardo.dsde-dev.broadinstitute.org/notebooks/aou-rw-test-3b102bbe/all-of-us-2278/notebooks/workspaces/qitest/qi-test.ipynb, Accept-Encoding: gzip, deflate, br, Accept-Language: en-US, en;q=0.9, zh-CN;q=0.8, zh;q=0.7, Cookie: <REDACTED>, X-Cloud-Trace-Context: 906e0fc6f601cdad3641d4fff58c6d5f/14103256683195023100, Via: 1.1 google, X-Forwarded-For: 69.173.127.109, 35.241.41.169, 35.191.2.128, 130.211.230.178, X-Forwarded-Proto: https, X-Forwarded-Host: leonardo.dsde-dev.broadinstitute.org, app:8080, X-Forwarded-Server: leonardo.dsde-dev.broadinstitute.org, all-of-us-2278-m.c.aou-rw-test-3b102bbe.internal, Content-Type: text/plain; charset=UTF-8, Connection: close, Content-Length: 47) body="{"localPath":"workspaces/qitest/qi-test.ipynb"}"
```


server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -122,6 +123,7 @@ class ObjectService(
         case None =>
           IO.raiseError(NotFoundException(s"${gsPath} not found in Google Storage"))
       }
+      _ <- permits.release
```
calbach: Lack of Scala knowledge: does this line get executed if you raise an error above, e.g. on line 124?

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -99,6 +99,7 @@ class ObjectService(
   //
   def acquireLock(req: AcquireLockRequest, traceId: TraceId): IO[Unit] =
     for {
+      _ <- permits.acquire
```
calbach: It feels too conservative to effectively set a global parallelism of 1 across the 3 primary methods in Welder; having multiple tabs is definitely a legitimate use case. I would recommend only acquiring locks per-notebook-path - via a map of semaphores or some other alternate library.

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
Qi77Qi: ah good point. It wouldn't

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -287,10 +287,21 @@ class ObjectService(
         IO.unit
       else IO.raiseError(InvalidLock("Fail to delocalize due to lock expiration or lock held by someone else"))
     } yield ()
+
+  private[server] def preventConcurrentAction[A](ioa: IO[A], localPath: RelativePath): IO[A] =
```
rtitle: This makes sense I think.
1. Get the Semaphore from the `Map[RelativePath, Semaphore[IO]]`
2. If the map entry is not present, create a new `Semaphore` and add it to the `Map`
3. Contruct an `IO` that `acquire`s the `Semaphore`, does the action, and `releases` the `Semaphore`

So this essentially implements a map of locks by `RelativePath`.

Qi77Qi: yep...`acquire` on `Semaphore` will semantically block until it becomes available again. The use of `Resource` guarantees `Semaphore` will be released even if ioa fails

PR #46, opened 2019-09-09
IA-1229

Qi77Qi:


server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -272,7 +272,12 @@ class ObjectService(
   private def delocalizeAndUpdateCache(localObjectPath: RelativePath, gsPath: GsPath, generation: Long, lock: Option[Lock], traceId: TraceId): IO[Unit] = {
     val lockMetadataToPush = lock.map(_.toMetadataMap).getOrElse(Map.empty)
     for {
-      delocalizeResp <- googleStorageAlg.delocalize(localObjectPath, gsPath, generation, lockMetadataToPush, traceId)
+      delocalizeResp <- googleStorageAlg.delocalize(localObjectPath, gsPath, generation, lockMetadataToPush, traceId).recoverWith {
+        case e: com.google.cloud.storage.StorageException if e.getCode == 412 =>
+          // In the case when the file is already been deleted from GCS, we try to delocalize the file with generation being 0L
```
calbach: This code path would also be hit if your version just goes out of sync normally as well. In other cases, we'd expect the second delocalize to fail as well (which is fine). Probably worth making this clearer in the comment.

PR #47, opened 2019-09-10
IA-1339 persist WORKSPACE_BUCKET

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-1339

Tested with Azmina's terra-ui branch

Screen Shot 2019-09-11 at 7 43 55 PM

### DIFF
--- a/server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/StorageLinksService.scala
+++ b/server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/StorageLinksService.scala
@@ -20,7 +20,12 @@ import _root_.io.chrisdavenport.linebacker.Linebacker
 import _root_.io.chrisdavenport.log4cats.Logger
 import org.broadinstitute.dsde.workbench.model.TraceId
 
-class StorageLinksService(storageLinks: StorageLinksCache, googleStorageAlg: GoogleStorageAlg, metadataCacheAlg: MetadataCacheAlg, config: StorageLinksServiceConfig)(
+class StorageLinksService(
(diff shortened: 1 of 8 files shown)


server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/StorageLinksService.scala
```diff
@@ -64,23 +77,32 @@ class StorageLinksService(storageLinks: StorageLinksCache, googleStorageAlg: Goo
     } yield link
   }
 
+  private def persistWorkspaceBucket(baseDirectory: LocalDirectory, cloudStorageDirectory: CloudStorageDirectory): IO[Unit] = {
+    val fileBody =
+      RuntimeVariables(s"gs://${cloudStorageDirectory.bucketName.value}/notebooks") //appending notebooks to mimick old jupyter image until we start using new images.
```
rtitle: This is to make it work with the current logic in `kernel-bootstrap.sh`?

Qi77Qi: yes

rtitle: Also I thought the old format of `.delocalize.json` was something like:
```
{
  "destination": "gs://foo/notebooks"
}
```
Does this output the same format?

rtitle: Ah nvm I see `runtimeVariablesEncoder` -- looks good.

PR #49, opened 2019-11-06
use structured logging

Qi77Qi:
Use mdc logging 
https://www.baeldung.com/mdc-in-log4j-2-logback

This also fixes a bug where welder is not updating local cache after acquiring lock


server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala
```diff
@@ -139,6 +139,7 @@ class ObjectService(
         updateMetadataResponse =>
           updateMetadataResponse match {
             case UpdateMetadataResponse.DirectMetadataUpdate => IO.unit
+              metadataCacheAlg.updateLock(localPath,lock)
```
Qi77Qi: this is the bug fix. Previously, when lock is updated properly in gcp, local cache for the lock is not updated properly

PR #52, opened 2020-02-05
[IA-1652] Keep less log files on disk

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-1652.

We stream log files to stackdriver anyway. We don't need to keep log files on disk for that long.


server/src/main/resources/logback.xml
```diff
@@ -21,7 +21,7 @@
         <maxFileSize>50MB</maxFileSize>
       </timeBasedFileNamingAndTriggeringPolicy>
       <!-- keep 30 days' worth of history -->
```
kyuksel:
```suggestion
      <!-- keep 3 days' worth of history -->
```

PR #53, opened 2020-02-11
[no ticket] scalafmt and dependency upgrade

Qi77Qi:


core/src/main/scala/org/broadinstitute/dsp/workbench/welder/Crc32c.scala
```diff
@@ -38,7 +38,7 @@ object Crc32c {
   private val BYTE_MASK = 0xff
 
   def calculateCrc32ForFile(path: java.nio.file.Path, blocker: Blocker)(implicit cs: ContextShift[IO]): IO[Crc32] =
-    fs2.io.file.readAll[cats.effect.IO](path, blocker, 4096).compile.to[Array].map(bytes => calculateCrc32c(bytes))
+    fs2.io.file.readAll[cats.effect.IO](path, blocker, 4096).compile.to(Array).map(bytes => calculateCrc32c(bytes))
```
rtitle: Is this an `fs2` change? I'm used to `to[A]`

Qi77Qi: yea...I think it's a new scala thing

Qi77Qi:
I made a few more minor updates like upgrading scalatest version to latest, and upgrading wb-libs to latest, and added a bit extra logging in `cacheResource`. Going to merge this as is, but let me know if there's more comments

PR #55, opened 2020-03-02
[IA-1709] Restart Welder if it crashes

akarukappadath:
Came out of [this spike](https://broadworkbench.atlassian.net/browse/IA-1452)

I manually tested this by creating a cluster, sshing into it, rewriting the `entrypoint.sh` on Welder, then killing and restarting the java process. I then overwhelmed Welder through a large number of localize requests and watched the logs to see the "respawning" message and saw Welder restarting. 

Now writes the output of `entrypoint.sh` into `/work/entrypoint.out` in the Welder container - this means that it is also in `/work/entrypoint.out` on the VM as well

### DIFF
--- a/project/Settings.scala
+++ b/project/Settings.scala
@@ -91,17 +91,10 @@ object Settings {
     )
 
   lazy val entrypoint = "/opt/docker/bin/entrypoint.sh"
+  lazy val entrypointLog = "/work/entrypoint.out"
 
   lazy val entrypointCmd =
-    s"""#!/bin/bash
-       |numask 002;
-       |until /opt/docker/bin/server; do
-       |  echo 'Welder crashed. Respawning..' >&2
-       |  sleep 1
-       |done""".stripMargin
-
-  lazy val entrypointTestCmd =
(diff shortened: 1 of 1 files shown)


project/Settings.scala
```diff
@@ -92,6 +92,14 @@ object Settings {
 
   lazy val entrypoint = "/opt/docker/bin/entrypoint.sh"
 
+  lazy val entrypointCmd =
+    s"""echo '#!/bin/bash\\numask 002;
+       |until /opt/docker/bin/server; do
+       |  echo 'Welder crashed. Respawning..' >&2
+       |  sleep 1
+       |done' > $entrypoint
```
Qi77Qi: maybe just replace this with `entrypoint`?
```suggestion
lazy val entrypointCmd =
    s"""echo '#!/bin/bash\\numask 002;
       |until /opt/docker/bin/server; do
       |  echo 'Welder crashed. Respawning..' >&2
       |  sleep 1
       |done' > /opt/docker/bin/entrypoint.sh
```

Qi77Qi:
nice! thanks for keep digging into this!

PR #56, opened 2020-03-24
IA-1707 upload metadata to gcs

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-1707

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsp/workbench/welder/package.scala
+++ b/core/src/main/scala/org/broadinstitute/dsp/workbench/welder/package.scala
@@ -11,13 +11,11 @@ import cats.effect.concurrent.Ref
 import cats.effect.{Blocker, ContextShift, IO, Resource, Sync}
 import cats.implicits._
 import fs2.{Pipe, Stream}
-import io.chrisdavenport.log4cats.Logger
 import io.circe.syntax._
 import io.circe.{Decoder, Encoder, Printer}
 import org.broadinstitute.dsde.workbench.google2.GcsBlobName
 import org.broadinstitute.dsde.workbench.model.WorkbenchEmail
 import org.broadinstitute.dsde.workbench.model.google.GcsBucketName
-import org.broadinstitute.dsde.workbench.util2
 import org.broadinstitute.dsp.workbench.welder.SourceUri.GsPath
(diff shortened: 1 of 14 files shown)


rtitle:
Just curious: have you tested this on a real cluster?

Is there any migration needed for existing clusters which currently have metadata on disk?

Qi77Qi:
the unit test should cover to some extent....but yea, I should test this in a cluster (https://broadworkbench.atlassian.net/browse/IA-1783)

For migration, that's a good point. Maybe I need to add logic when starting up welder, if the file exists on disk, then it should load from disk; otherwise it gets it from gcp

Qi77Qi:
I tested things out in fiab with an older runtime, and then updated leonardo's welder hash. The upgrade script runs successfully, `.welder` is being loaded correctly, and then deleted by background process.

Those cache files are sent to gcs properly as well.
Screen Shot 2020-03-26 at 3 31 16 PM

Started and stopped the runtime a few more times, and things work fine.

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/package.scala
```diff
@@ -111,32 +113,71 @@ package object welder {
...
+    for {
+      data <- Stream.eval(ref.get)
+      _ <- googleStorageAlg.uploadBlob(stagingBucketName, blobName, data.values.toSet.asJson.printWith(Printer.noSpaces).getBytes("UTF-8"))
+      // We're previously reading and persisting cache from/to local disk, but this can be problematic when disk space runs out.
+      // Hence we're persisting cache to GCS. Since leonardo tries to automatically upgrade welder version, we'll need to support both cases.
+      // We first try to read cache from local disk, if it exists, use it; if not, we read cache from gcs
+      // Code for reading from disk can be deleted once we're positive that all user clusters are upgraded or when we no longer care.
+      // This change is made on 3/26/2020
+      legacyCacheDir = new Directory(new File(s"/work/.welder"))
+      _ <- Stream.eval(IO(legacyCacheDir.deleteRecursively()))
```
Qi77Qi: if the directory doesn't exist, this returns `false`, which we don't care here

PR #68, opened 2021-08-25
[IA-2653] Background process for syncing Rmd files

gpcarr:
This PR adds a background process to welder that will only run on RStudio runtimes every 15 seconds. The background process checks if there are changes in the file before delocalizing it instead of just blindly delocalizing every 15 seconds. It also makes `localSafeModeBaseDirectory` optional in the storageLinks API since we will not need this field for Rmd files. 

Tested with both unit tests and running welder locally. When running welder locally, validated that a `.Rmd` file that lives on my machine gets delocalized to my workspace bucket. 

This is the accompanying Leo PR: https://github.com/DataBiosphere/leonardo/pull/2259 where the `IS_RSTUDIO_RUNTIME` env var is set


Qi77Qi:
one thing just occurred to me, I think it can potentially be a lot of traffic if we sync periodically blindly. But instead, we can check if the file has been updated before delocalizing, similar to [this](https://github.com/DataBiosphere/welder/blob/master/server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala#L172)...basically calculates `crc32c` for the file, and compare it to the file's crc32c in google if it exists, and only delocalize if this hash has changed, or the file doesn't exist in google yet

kyuksel:
> one thing just occurred to me, I think it can potentially be a lot of traffic if we sync periodically blindly. But instead, we can check if the file has been updated before delocalizing, similar to this...basically calculates crc32c for the file, and compare it to the file's crc32c in google if it exists, and only delocalize if this hash has changed, or the file doesn't exist in google yet

Sounds like a great idea  👍🏼

gpcarr:
> one thing just occurred to me, I think it can potentially be a lot of traffic if we sync periodically blindly. But instead, we can check if the file has been updated before delocalizing, similar to [this](https://github.com/DataBiosphere/welder/blob/master/server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/ObjectService.scala#L172)...basically calculates `crc32c` for the file, and compare it to the file's crc32c in google if it exists, and only delocalize if this hash has changed, or the file doesn't exist in google yet

I like this idea too, I'll update the code and test!

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/BackgroundTask.scala
```diff
@@ -79,12 +83,68 @@ class BackgroundTask(
...
+          } else IO.unit
+        }
+      } yield ()
+      (Stream.sleep[IO](2 seconds) ++ Stream.eval(res)).repeat
+    } else {
+      Stream.eval(logger.info("Not running rmd sync process because this is not an Rstudio runtime"))
+    }
+  }
+
+  def shouldDelocalize(gsPath: GsPath, localAbsolutePath: Path): IO[Boolean] =
```
Qi77Qi, line 119: can you add a unit test for this...mocking out the `getObjectMetadata`'s response

server/src/test/resources/reference.conf
```diff
@@ -1 +1,5 @@
-working-directory = "/tmp"
\ No newline at end of file
+object-service {
```
Qi77Qi, line 1: I don't think these are needed...we need to test configs are reading environment variables properly too...

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/BackgroundTask.scala
gpcarr: just pushed the 2 unit tests!

core/src/test/scala/org/broadinstitute/dsp/workbench/welder/BackgroundTaskSpec.scala
```diff
@@ -0,0 +1,77 @@
...
+
+  "getGsPath" should "return the correct path to delocalize files to" in {
+    val storageLink = genRmdStorageLink.sample.get
+    val file = new File("test.Rmd")
+    val res = initBackgroundTask(Map(storageLink.localBaseDirectory.path -> storageLink), Map.empty, None, blocker).getGsPath(storageLink, file)
+    res.toString shouldBe s"gs://${storageLink.cloudStorageDirectory.bucketName.value}/${storageLink.cloudStorageDirectory.blobPath.get.asString}/test.Rmd"
+  }
+
+  "shouldDelocalize" should "return true if files have changed" in {
+    val metadataResp = GetMetadataResponse.Metadata(Crc32("aZKdIw=="), Map.empty, 0L) //This crc32c is from gsutil
```
Qi77Qi, line 27: think the comment is not valid?

gpcarr: true - will update

server/src/main/resources/reference.conf
```diff
@@ -11,3 +11,5 @@ staging-bucket-name = ${STAGING_BUCKET}
 clean-up-lock-interval = 7 minutes
 flush-cache-interval = 10 minutes
 sync-cloud-storage-directory-interval = 15 minutes
+delocalize-directory-interval = 15 seconds
```
Qi77Qi, line 14: maybe every 60 seconds? every 15 seconds feels really frequent 🤔

gpcarr: hmm I did every 15 as a suggestion from @rtitle but am open to discussion on how frequently we want to delocalize

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/BackgroundTask.scala
```diff
@@ -79,12 +83,68 @@ class BackgroundTask(
...
+                } else {
+                  IO.unit
+                }
+              } yield ()
+
+            }
+          } else IO.unit
+        }
+      } yield ()
+      (Stream.sleep[IO](2 seconds) ++ Stream.eval(res)).repeat
```
Qi77Qi, line 113: this duration needs to be changed to what's in config

server/src/main/resources/logback.xml
```diff
@@ -27,10 +27,12 @@
 
   <logger name="org.broadinstitute.dsp.workbench.welder" level="INFO" additivity="false">
     <appender-ref ref="FILE"/>
+    <appender-ref ref="CONSOLE"/>
```
Qi77Qi, line 30: revert this change

Qi77Qi:
```[info] shouldDelocalize
[info] - should return true if files have changed *** FAILED ***
[info]   false was not equal to true (BackgroundTaskSpec.scala:37)
[info] shouldDelocalize
[info] - should return false if files have not changed *** FAILED ***
[info]   true was not equal to false (BackgroundTaskSpec.scala:52)
```
new unit tests are failing

Also, can you add one more that verifies when there's the file doens't exist in google,`shouldDelocalize` returns `true`

gpcarr:
> ```
> [info] - should return true if files have changed *** FAILED ***
> [info]   false was not equal to true (BackgroundTaskSpec.scala:37)
> [info] shouldDelocalize
> [info] - should return false if files have not changed *** FAILED ***
> [info]   true was not equal to false (BackgroundTaskSpec.scala:52)
> ```
> 
> new unit tests are failing
> 
> Also, can you add one more that verifies when there's the file doens't exist in google,`shouldDelocalize` returns `true`

yeah I'll add that additional test. that's weird that they're failing here, they're passing locally. i'll look into it

server/src/main/resources/logback.xml
```diff
@@ -31,6 +31,7 @@
 
   <logger name="org.http4s.server.blaze.Http1ServerStage" level="INFO" additivity="false">
     <appender-ref ref="FILE"/>
+    <appender-ref ref="CONSOLE"/>
```
Qi77Qi, line 34: revert this change

core/src/test/scala/org/broadinstitute/dsp/workbench/welder/mock.scala
```diff
@@ -0,0 +1,22 @@
+package org.broadinstitute.dsp.workbench.welder
```
Qi77Qi, line 1: it's probably my fault...but can you name this file to `mocks`

Qi77Qi (approved):
:shipit:  🎉

PR #70, opened 2021-08-30
IA-2876 use blessed image

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2876

### DIFF
--- a/project/Settings.scala
+++ b/project/Settings.scala
@@ -76,7 +76,7 @@ object Settings {
 
   lazy val commonDockerSettings = List(
     maintainer := "workbench-interactive-analysis@broadinstitute.org",
-    dockerBaseImage := "us.gcr.io/broad-dsp-gcr-public/base/jre:11-debian",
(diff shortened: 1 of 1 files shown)


rtitle:
🎉 

Wondering if we need `debian`? I think Leo needed it for `helm` stuff but wondering if welder could use `alpine` or even `distroless`: https://console.cloud.google.com/gcr/images/broad-dsp-gcr-public/us/base/jre?project=broad-dsp-gcr-public

Qi77Qi:
oh good point..let me try the other two

Qi77Qi:
@rtitle tried both distroless and alpine...both failed because either `bash` doesn't exist or `/bin/sh` doesn't exist...so I'm going to keep it as `debian` for now...the it's abt 30mg size difference between `debian` and `alpine`

kyuksel (approved):

rtitle:
Ah ok 👍

rtitle (approved):

PR #71, opened 2021-08-31
IA-2876 add cloud build for publishing welder

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2876

similar to https://github.com/broadinstitute/leonardo-cron-jobs/blob/master/resource-validator/cloudbuild.yaml


rtitle (approved):
LGTM if it works :)

PR #74, opened 2022-01-21
[IA-2991] Rmd file syncing

gpcarr:
delocalization looks good locally - creating a file on my local machine, hitting the `POST storageLinks` endpoint and validating that the file gets delocalized to bucket looks good.

Tested end to end in a Fiab with this Leo [PR](https://github.com/DataBiosphere/leonardo/pull/2530) and this UI [PR](https://github.com/DataBiosphere/terra-ui/pull/2810) and everything works as expected. Welder throws a generation mismatch error if user1 creates file1, then file1 gets localized to user2's VM, then user2 makes some changes. then user1 comes back and does not get user2's changes and makes their own changes and tries to save -> when user1 saves, they get a generation mismatch error since they have an outdated version of file1. verified that file1's metadata gets updated in GCS when this happens to reflect that they have an outdated version of the file.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsp/workbench/welder/BackgroundTask.scala
+++ b/core/src/main/scala/org/broadinstitute/dsp/workbench/welder/BackgroundTask.scala
@@ -8,8 +8,8 @@ import cats.implicits._
 import cats.mtl.Ask
 import fs2.Stream
 import org.typelevel.log4cats.Logger
-import org.broadinstitute.dsde.workbench.google2.{GcsBlobName, GetMetadataResponse, GoogleStorageService}
-import org.broadinstitute.dsde.workbench.model.TraceId
+import org.broadinstitute.dsde.workbench.google2.GcsBlobName
+import org.broadinstitute.dsde.workbench.model.{TraceId, WorkbenchEmail}
 import org.broadinstitute.dsde.workbench.model.google.GcsBucketName
 import org.broadinstitute.dsp.workbench.welder.JsonCodec._
 import org.broadinstitute.dsp.workbench.welder.SourceUri.GsPath
@@ -23,7 +23,6 @@ class BackgroundTask(
     storageLinksCache: StorageLinksCache,
     googleStorageAlg: GoogleStorageAlg,
     metadataCacheAlg: MetadataCacheAlg,
-    googleStorageService: GoogleStorageService[IO],
     blocker: Blocker
 )(implicit cs: ContextShift[IO], logger: Logger[IO], timer: Timer[IO]) {
   val cleanUpLock: Stream[IO, Unit] = {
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/broadinstitute/dsp/workbench/welder/BackgroundTask.scala
```diff
@@ -116,18 +103,54 @@ class BackgroundTask(
...
-      case Some(GetMetadataResponse.Metadata(crc32, _, _)) =>
-        if (localCrc32c == crc32)
-          false
-        else
-          true
-      case _ => true
-    }
+      traceId <- ev.ask[TraceId]
+      localAbsolutePath = config.workingDirectory.resolve(localObjectPath.asPath) //TODO: make sure this is correct
+      _ <- logger.info(s"local absolute path ${localAbsolutePath}")
```
Qi77Qi, line 110: just reminder to remove this before merge

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/BackgroundTask.scala
```diff
@@ -116,18 +103,54 @@ class BackgroundTask(
...
+            else {
+              // In the case when the file is already been deleted from GCS, we try to delocalize the file with generation being 0L
+              // This assumes the business logic we want is always to recreate files that have been deleted from GCS by other users.
+              // If the file is indeed out of sync with remote, both delocalize attempts will fail due to generation mismatch
+              googleStorageAlg.delocalize(localObjectPath, gsPath, 0L, lastModifiedByMetadataToPush, traceId)
+            }
+        }
+        .recoverWith {
+          case e: com.google.cloud.storage.StorageException if e.getCode == 412 =>
+            googleStorageAlg.delocalize(localObjectPath, gsPath, 0L, Map("ownerEmail" -> "outdated"), traceId)
```
Qi77Qi, line 149: I don't think we should delocalize again. This will just fail again with 412...We need to make a call that just updates metadata...also `ownerEmail` needs to be populated properly

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/BackgroundTask.scala
```diff
@@ -86,48 +85,72 @@ class BackgroundTask(
...
-      case Some(GetMetadataResponse.Metadata(crc32, _, _)) =>
-        if (localCrc32c == crc32)
-          false
-        else
-          true
-      case _ => true
-    }
+      traceId <- ev.ask[TraceId]
+      localAbsolutePath = config.workingDirectory.resolve(localObjectPath.asPath) //TODO: make sure this is correct
+      _ <- logger.info(s"! local absolute path ${localAbsolutePath}")
```
Qi77Qi, line 110: remove this log line

Qi77Qi (dismissed):
LGTM! can you update the wiki doc in terms of how `outdated` field will be used in UI, and when the field will be cleaned up as well

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/GoogleStorageInterp.scala
```diff
@@ -28,7 +28,7 @@ class GoogleStorageInterp(config: GoogleStorageAlgConfig, blocker: Blocker, goog
   private val chunkSize = 1024 * 1024 * 2 // com.google.cloud.storage.BlobReadChannel.DEFAULT_CHUNK_SIZE
 
   def updateMetadata(gsPath: GsPath, traceId: TraceId, metadata: Map[String, String]): IO[UpdateMetadataResponse] =
-    googleStorageService
+    logger.info(s"updateMetadata: ${metadata}") >> googleStorageService
```
Qi77Qi, line 31: remove the extra logging...google2 logs the setMetadata call

Qi77Qi (changes requested):
one small comment

Qi77Qi (approved):

PR #75, opened 2022-02-25
[no ticket] upgrade to cats-effect 3

Qi77Qi:
while investigating https://broadworkbench.atlassian.net/browse/IA-3222, I notice welder hasn't been updated for a while in terms of dependencies. So this PR mostly just upgrade it to use cats-effect 3


xuscrbw (approved):
LGTM

PR #76, opened 2022-03-23
[IA-3209] Handle when a user does not want their file synced

gpcarr:
See corresponding UI PR for testing https://github.com/DataBiosphere/terra-ui/pull/2810

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsp/workbench/welder/BackgroundTask.scala
+++ b/core/src/main/scala/org/broadinstitute/dsp/workbench/welder/BackgroundTask.scala
@@ -92,7 +92,7 @@ class BackgroundTask(
           if (storageLink.pattern.toString.contains(".Rmd")) {
             findFilesWithPattern(config.workingDirectory.resolve(storageLink.localBaseDirectory.path.asPath), storageLink.pattern).traverse_ { file =>
               val gsPath = getGsPath(storageLink, new File(file.getName))
-              safeDelocalize(
+              checkSyncStatus(
                 gsPath,
                 RelativePath(java.nio.file.Paths.get(file.getName))
               )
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/broadinstitute/dsp/workbench/welder/GoogleStorageInterp.scala
```diff
@@ -56,6 +56,17 @@ class GoogleStorageInterp(config: GoogleStorageAlgConfig, blocker: Blocker, goog
       }
     } yield res
 
+  def retrieveUserDefinedMetadata(gsPath: GsPath, traceId: TraceId): IO[Option[Map[String, String]]] =
```
Qi77Qi, line 59: `Map` can be empty...I would just return `IO[Map[String, String]]`

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/BackgroundTask.scala
```diff
@@ -106,7 +106,24 @@ class BackgroundTask(
...
 
-  def safeDelocalize(gsPath: GsPath, localObjectPath: RelativePath)(implicit ev: Ask[IO, TraceId]): IO[Unit] =
+  def checkSyncStatus(gsPath: GsPath, localObjectPath: RelativePath)(implicit ev: Ask[IO, TraceId]): IO[Unit] =
+    for {
+      traceId <- ev.ask[TraceId]
+      hashedOwnerEmail <- IO.fromEither(hashString(config.ownerEmail.value))
+      bucketMetadata <- googleStorageAlg.retrieveUserDefinedMetadata(gsPath, traceId)
+      _ <- bucketMetadata match {
+        case Some(meta) => {
+          if (meta.keySet.contains(hashedOwnerEmail.asString) && meta.get(hashedOwnerEmail.asString).getOrElse(None) == "doNotSync") {
```
Qi77Qi, line 116: `None` is not a `String`....`meta.get(hashedOwnerEmail.asString).getOrElse(None) == "doNotSync"` is comparing `None` and `doNotSync`, which will always be `false`...it happens to work fine in this case...but this is not great.

`meta.get(hashedOwnerEmail.asString).contains("doNotSync")` will be better...also some comments about where this `doNotSync` comes from will be useful

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/BackgroundTask.scala
```diff
@@ -106,7 +106,26 @@ class BackgroundTask(
     }
   }
 
-  def safeDelocalize(gsPath: GsPath, localObjectPath: RelativePath)(implicit ev: Ask[IO, TraceId]): IO[Unit] =
+  def checkSyncStatus(gsPath: GsPath, localObjectPath: RelativePath)(implicit ev: Ask[IO, TraceId]): IO[Unit] =
+    for {
+      traceId <- ev.ask[TraceId]
+      hashedOwnerEmail <- IO.fromEither(hashString(config.ownerEmail.value))
+      bucketMetadata <- googleStorageAlg.retrieveUserDefinedMetadata(gsPath, traceId)
+      _ <- logger.info("bucketMetadata: " + bucketMetadata.get(hashedOwnerEmail.asString) + " email: " + hashedOwnerEmail.asString)
```
Qi77Qi, line 114: can you remove these extra debug lines

Qi77Qi (approved):
lgtm..

One thing to confirm is that if the file hasn't changed, we won't be delocaizing it.

PR #79, opened 2022-04-20
SUP-643  bump a few dependencies

Qi77Qi:
We're still seeing outdated cache issue. Make flushing even more frequent and bump a few dependencies for now


xuscrbw (approved):
lgtm

PR #80, opened 2022-05-25
[IA-3445] Added .R file support

cstrandw:
Added .R file support so that .R files will be delocalized.

Testing steps:
- Created a local Welder server
- Created .R files on my computer
- Made a POST request to the server so that it would delocalize .R files (ie. `curl -vX POST --header 'Content-Type: application/json' --header 'Accept: application/json' localhost:8080/storageLinks -d '{"localBaseDirectory": "", "localSafeModeBaseDirectory": "", "cloudStorageDirectory": "gs://fc-d01f8619-06cd-443d-86f2-8e80b0fb33f3/notebooks", "pattern": ".*\\.R" }'`)
- Verified that the files delocalized to the workspace bucket for the given workspace
- Verified that edits to a file also delocalized as expected
- Repeated these steps for .Rmd files as well to make sure support was not deprecated by my changes

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsp/workbench/welder/BackgroundTask.scala
+++ b/core/src/main/scala/org/broadinstitute/dsp/workbench/welder/BackgroundTask.scala
@@ -92,7 +92,7 @@ class BackgroundTask(
         storageLinks <- storageLinksCache.get
         implicit0(tid: Ask[IO, TraceId]) <- IO(TraceId(UUID.randomUUID().toString)).map(tid => Ask.const[IO, TraceId](tid))
         _ <- storageLinks.values.toList.traverse { storageLink =>
-          if (storageLink.pattern.toString.contains(".Rmd")) {
+          if (storageLink.pattern.toString == ".*\\.Rmd" || storageLink.pattern.toString == ".*\\.R") {
             findFilesWithPattern(config.workingDirectory.resolve(storageLink.localBaseDirectory.path.asPath), storageLink.pattern).traverse_ { file =>
               val gsPath = getGsPath(storageLink, new File(file.getName))
               checkSyncStatus(


core/src/main/scala/org/broadinstitute/dsp/workbench/welder/BackgroundTask.scala
```diff
@@ -92,7 +92,7 @@ class BackgroundTask(
         storageLinks <- storageLinksCache.get
         implicit0(tid: Ask[IO, TraceId]) <- IO(TraceId(UUID.randomUUID().toString)).map(tid => Ask.const[IO, TraceId](tid))
         _ <- storageLinks.values.toList.traverse { storageLink =>
-          if (storageLink.pattern.toString.contains(".Rmd")) {
+          if (storageLink.pattern.toString == ".*\\.Rmd" || storageLink.pattern.toString == ".*\\.R") {
```
Qi77Qi, line 95: maybe we can have a function like
```
val filesShouldSyncRegex = ".*(.R)$|(.Rmd)$".r //put this on top of the class
def shouldSync(storageLinkPattern String): Boolean = filesShouldSyncRegex.findFirstIn(storageLinkPattern)
```

and use this function here instead.

Also add some unit tests to test out the function

core/src/test/scala/org/broadinstitute/dsp/workbench/welder/BackgroundTaskSpec.scala
```diff
@@ -37,6 +37,18 @@ class BackgroundTaskSpec extends AnyFlatSpec with WelderTestSuite {
...
+  "shouldSync" should "return true if passed a supported pattern" in {
+    val storageLink =
+      StorageLink(
+        LocalBaseDirectory(RelativePath(Paths.get(""))),
+        None,
+        CloudStorageDirectory(GcsBucketName("testBucket"), Some(BlobPath("notebooks"))),
+        ".*\\.R".r
+      )
+    val res = initBackgroundTask(Map(storageLink.localBaseDirectory.path -> storageLink), Map.empty, None).shouldSync(storageLink.pattern.toString())
+    res shouldBe true
```
Qi77Qi, line 49: this looks pretty good..just realize BackgroundTask is a class....to make testing a bit easier, I would put `shouldSync` in `BackgroundTask` object.

So in `[BackgroundTask.scala](https://github.com/DataBiosphere/welder/pull/80/files#diff-4a38dd3f6392b9fe795b64ff8746380bb5c0ce340986c04ce2dc0bcb3ac8207e)` file,

```
object BackgroundTask {
  val filesShouldSyncRegex: Regex = ".*(.R)$|(.Rmd)$".r
  def shouldSync(storageLinkPattern: String): Boolean = filesShouldSyncRegex.findFirstIn(storageLinkPattern).isDefined
}  
```


This way, in test, you can just do `BackgroundTask.shouldSync` to access it.

in I would add a few more tests
like
```
BackgroundTask.shouldSync("da.R") shouldBe true
BackgroundTask.shouldSync("da.Rdd") shouldBe false
BackgroundTask.shouldSync("da.Rmd") shouldBe true
...
```

core/src/test/scala/org/broadinstitute/dsp/workbench/welder/BackgroundTaskSpec.scala
```diff
@@ -37,6 +37,13 @@ class BackgroundTaskSpec extends AnyFlatSpec with WelderTestSuite {
     res.toString shouldBe s"gs://testBucket/notebooks/test.Rmd"
   }
```
Qi77Qi, line 39: can you add one more test `test.Rmddd`

Qi77Qi (approved):
looks great!

core/src/test/scala/org/broadinstitute/dsp/workbench/welder/BackgroundTaskSpec.scala
Qi77Qi: no...that's ok...it's nice to have it be green..but we haven't been very strict about it

Qi77Qi (approved):

PR #84, opened 2022-07-16
PROD-695 better fix

Qi77Qi:
This reverts commit 99d2ea7c7919b8d9adaf5ec481d01b7dd283a57b.

So in previous PR, we were hard coding the pattern, but I think shouldn't need to. I was previously confused thinking we shouldn't sync ipynb files, hence we need to specifically filter on .R or .Rmd files, but we're already guarding that with the config check to see if this is an RStudio runtime.

### DIFF
--- a/core/src/test/scala/org/broadinstitute/dsp/workbench/welder/PackageSpec.scala
+++ b/core/src/test/scala/org/broadinstitute/dsp/workbench/welder/PackageSpec.scala
@@ -79,13 +79,16 @@ class PackageSpec extends AnyFlatSpec with ScalaCheckPropertyChecks with WelderT
     }
   }
 
-  "findFilesWithROrRmdPattern" should "only find files with certain pattern" in {
+  "findFilesWithPattern" should "only find files with certain pattern" in {
     val parentPath = Paths.get("/tmp")
     new java.io.File(parentPath.resolve("test.R").toUri).createNewFile()
     new java.io.File(parentPath.resolve("test.R1").toUri).createNewFile()
     new java.io.File(parentPath.resolve("test.Rmd").toUri).createNewFile()
     new java.io.File(parentPath.resolve("test.ipynb").toUri).createNewFile()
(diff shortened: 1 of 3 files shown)


core/src/test/scala/org/broadinstitute/dsp/workbench/welder/PackageSpec.scala
```diff
@@ -79,13 +79,16 @@ class PackageSpec extends AnyFlatSpec with ScalaCheckPropertyChecks with WelderT
...
 
-  "findFilesWithROrRmdPattern" should "only find files with certain pattern" in {
+  "findFilesWithPattern" should "only find files with certain pattern" in {
     val parentPath = Paths.get("/tmp")
     new java.io.File(parentPath.resolve("test.R").toUri).createNewFile()
     new java.io.File(parentPath.resolve("test.R1").toUri).createNewFile()
     new java.io.File(parentPath.resolve("test.Rmd").toUri).createNewFile()
     new java.io.File(parentPath.resolve("test.ipynb").toUri).createNewFile()
-    val resR = findFilesWithROrRmdPattern(parentPath)
+    val resR = findFilesWithPattern(parentPath, ".*(.R)$|(.Rmd)$".r)
```
gpcarr, line 88: This change looks good and makes sense - this makes Welder independent of needing to know any regex for an application. The only thing to confirm now is that the UI is sending the right regex which I think would be ".R(md)?" (The UI is currently sending this pattern without a question mark which is not correct) Or something like that, I can't remember off the top of my head. However, this unit test is testing a different regex pattern, is the plan to have the UI send `".*(.R)$|(.Rmd)$".r`?

Qi77Qi: yes, we’ll need to confirm the regex in UI is correct with this change
updating the regex (if needed) should be relatively easy

gpcarr (approved):
:shipit:

PR #88, opened 2022-08-08
[IA-3511] Add azure storage alg

jdcanas:


core/src/main/scala/org/broadinstitute/dsp/workbench/welder/model.scala
```diff
@@ -45,11 +46,42 @@ object SyncMode {
...
+  final case class DataUri(data: Array[Byte]) extends SourceUri {
+    override val cloudProvider: CloudProvider = CloudProvider.None
+  }
   final case class GsPath(bucketName: GcsBucketName, blobName: GcsBlobName) extends SourceUri {
     override def toString: String = s"gs://${bucketName.value}/${blobName.value}"
+
+    override val cloudProvider: CloudProvider = CloudProvider.Gcp
+  }
+  final case class AzurePath(containerName: ContainerName, blobName: BlobName) extends SourceUri {
+    override def toString: String = s"gs://${containerName.value}/${blobName.value}"
```
jdcanas, line 82: fix

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/CloudStorageAlg.scala
```diff
@@ -9,41 +9,52 @@ import org.broadinstitute.dsde.workbench.model.TraceId
...
 import scala.util.matching.Regex
 
 trait CloudStorageAlg {
-  def updateMetadata(gsPath: GsPath, traceId: TraceId, metadata: Map[String, String]): IO[UpdateMetadataResponse]
-  def retrieveAdaptedGcsMetadata(localPath: RelativePath, gsPath: GsPath, traceId: TraceId): IO[Option[AdaptedGcsMetadata]]
-  def retrieveUserDefinedMetadata(gsPath: GsPath, traceId: TraceId): IO[Map[String, String]]
-  def removeObject(gsPath: GsPath, traceId: TraceId, generation: Option[Long]): Stream[IO, RemoveObjectResult]
+  def cloudProvider: CloudProvider
+
+  abstract def updateMetadata(gsPath: SourceUri, traceId: TraceId, metadata: Map[String, String]): IO[UpdateMetadataResponse] =
```
jdcanas, line 21: abstract modifier probably not needed

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/CloudStorageAlg.scala
```diff
@@ -52,17 +63,19 @@ trait CloudStorageAlg {
...
-  def localizeCloudDirectory(
+  abstract  def localizeCloudDirectory(
       localBaseDirectory: RelativePath,
       cloudStorageDirectory: CloudStorageDirectory,
       workingDir: Path,
       pattern: Regex,
       traceId: TraceId
-  ): Stream[IO, AdaptedGcsMetadataCache]
+  )(implicit ev: Ask[IO, TraceId]): Stream[IO, Option[AdaptedGcsMetadataCache]] =
+    Stream.eval(IO.raiseError(new RuntimeException(s"$cloudProvider storage interp localizeCloudDirectory should only be called with appropriate source URI")))
```
jdcanas, line 73: RuntimeException shouldn't be used, WIP

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/AzureStorageInterp.scala
```diff
@@ -61,9 +108,50 @@ class AzureStorageInterp(implicit logger: StructuredLogger[IO]) extends CloudSto
...
+        case Some(blobName) =>
+          for {
+            localPath <- Stream.eval(IO.fromEither(getLocalPath(localBaseDirectory, cloudStorageDirectory.blobPath, blob.getName)))
+            localAbsolutePath <- Stream.fromEither[IO](Either.catchNonFatal(workingDir.resolve(localPath.asPath)))
+
+            result <- localAbsolutePath.toFile.exists() match {
+              case true => Stream(None).covary[IO]
+              case false =>
+                Stream.eval(mkdirIfNotExist(localAbsolutePath.getParent)) >>
+                  gcsToLocalFile(localAbsolutePath, AzurePath(azureContainer, BlobName(blobName)), traceId)
```
jdcanas, line 133: this is where this implementation differs slightly from the googleStorageInterp, here we call the interface method `gcsToLocalFile` that exists within the welder azureStorageinterp already instead of a built-in function on the cloud sdk blob object

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/package.scala
```diff
@@ -32,8 +33,10 @@ package object server {
           miscHttpClient = new MiscHttpClientInterp(client, conf.miscHttpClientConfig)
           petAccessTokenResp <- Resource.eval(miscHttpClient.getPetAccessToken())
           sasTokenResp <- Resource.eval(miscHttpClient.getSasUrl(petAccessTokenResp.accessToken))
-          //TODO: Justin
-        } yield ???
+          //TODO: transform sas token resp into tok[en and url
```
Qi77Qi, line 36: this TODO still needs to be addressed?

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/AzureStorageInterp.scala
```diff
@@ -59,11 +113,43 @@ class AzureStorageInterp(implicit logger: StructuredLogger[IO]) extends CloudSto
...
+  override def uploadBlob(path: SourceUri)(implicit ev: Ask[IO, TraceId]): Pipe[IO, Byte, Unit] = path match {
+    case AzurePath(containerName, blobName) => azureStorageService.uploadBlob(containerName, blobName)
+    case _ => super.uploadBlob(path)
+  }
 
-  override def uploadBlob(bucketName: GcsBucketName, objectName: GcsBlobName): Pipe[IO, Byte, Unit] = ???
+  override def getBlob[A: Decoder](path: SourceUri)(implicit ev: Ask[IO, TraceId]): fs2.Stream[IO, A] =
+    for {
+      traceId <- Stream.eval(ev.ask)
+      error <- Stream.eval(IO.raiseError[A](NotImplementedException(traceId, "getBlob not implemented for azure storage interp")))
```
Qi77Qi, line 152: this needs to be implemented

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/package.scala
```diff
@@ -129,17 +128,34 @@ package object welder {
...
+            loadedCache <- cacheFromDisk.fold {
+              storageAlg
+                .getBlob[List[B]](sourceUri)
+                .last
+                .map(_.getOrElse(List.empty)) // The first time welder starts up, there won't be any existing cache, hence returning empty list
+            }(x => Stream.eval(IO.pure(x)))
+          } yield loadedCache
+        case SourceUri.AzurePath(containerName, blobName) =>
+          for {
+            cacheFromDisk <- localCache[B](Paths.get(s"/work/.welder/${blobName.value.split("/")(1)}"))
```
Qi77Qi, line 146:
```
localCache[B](Paths.get(s"/work/.welder/${blobName.value.split("/")(1)}"))
```
can be removed...it's not needed since this was just for backwards compatibility, and we don't need to worry about that for Azure right now

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/model.scala
```diff
@@ -45,11 +46,42 @@ object SyncMode {
...
+
+  final case object Gcp extends CloudProvider {
+    override val asString = "GCP"
+  }
+
+  final case object Azure extends CloudProvider {
+    override val asString = "AZURE"
+  }
+
+  final case object None extends CloudProvider {
```
Qi77Qi, line 62: what is this for? can we remove this?

jdcanas: This is needed for the sourceUri contract, as a sourceURI must have a cloudProvider. A DataURI is a type of sourceURI that does not explicitly have a cloud provider, so we need this enum value to ensure consumers of a `SourceURI` handle those that are not azure/gcp specific properly.

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/Main.scala
```diff
@@ -48,8 +53,7 @@ object Main extends IOApp {
       )
       metadataCache <- cachedResource[RelativePath, AdaptedGcsMetadataCache](
         storageAlgRef,
-        appConfig.stagingBucketName,
-        appConfig.metadataJsonBlobName,
+        appConfig.getSourceUri,
```
Qi77Qi, line 56: this is wrong...the blob path for storagelink and metadata cache is different

jdcanas: think I fixed it properly, with the assumption this is only in use for google currently...

core/src/main/scala/org/broadinstitute/dsp/workbench/welder/model.scala
Qi77Qi: feels like we probably can model things a bit better tho...It doesn't feel right to add an arbitrary case becuz call site of `CloudProvider` has a specific use case to satisfy.

But we can leave this as is for this PR

server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/Main.scala
Qi77Qi: it'll be used for azure too..but I'll fix in my follow up PR

Qi77Qi (approved):
think we can merge this for now...there're still some issues for Azure case, but I'll address in my PR

PR #114, opened 2022-08-29
IA-3655 Make it work for Azure

Qi77Qi:
This PR makes syncing ipynb files work along with https://github.com/DataBiosphere/leonardo/pull/2818

Related wb-libs PR https://github.com/broadinstitute/workbench-libs/pull/1074


core/src/main/scala/org/broadinstitute/dsp/workbench/welder/package.scala
```diff
@@ -204,16 +197,22 @@ package object welder {
...
-      implicit ev: Ask[IO, TraceId]
-  ): IO[SourceUri] =
-    ev.ask[TraceId].flatMap { traceId =>
-      cloudProvider match {
-        case CloudProvider.Gcp => IO(GsPath(container.asGcsBucket, blob.asGcs))
-        case CloudProvider.Azure => IO(AzurePath(container.asAzureCloudContainer, blob.asAzure))
-        case CloudProvider.None => IO.raiseError(InvalidSourceURIException(traceId, "Cannot get sourceURI with no cloud provider", Map.empty))
-      }
-    }
+//  def getSourceUriForProvider(cloudProvider: CloudProvider, container: CloudStorageContainer, blob: CloudStorageBlob)(
```
jdcanas, line 200: Seems like this helper is probably not needed since we don't transform a `CloudStorageContainer` +`CloudStorageBlob` + `CloudProvider` into a `SourceUri` now, and use the the new types instead?

Qi77Qi: ah yea...will clean up

server/src/test/resources/reference.conf
```diff
@@ -1,8 +1,10 @@
...
 is-rstudio-runtime = false
 type = "azure"
 misc-http-client-config = {
   wsm-url = "https://workspace.dsde-dev.broadinstitute.org/"
   workspace-id = "a5a1f1e1-bcb0-49d9-b589-ea4d7c9d6f02"
-  storage-container-resource-id = "9151f3a0-fe5c-49c5-b8a1-dc15cd96b174"
-}
\ No newline at end of file
+}
+workspace-storage-container-resource-id = "9151f3a0-fe5c-49c5-b8a1-dc15cd96b174"
```
jdcanas, line 9: are these stable entities that exist in the cloud and tests depend on, or dummy values that are never hit because the relevant calls are mocked? Probably worth making a comment to indicate which in the file.

jdcanas (dismissed):
Changing these types cascades into a bunch of changes. Think I spotted most of the actual logic changes and commented where applicable, overall looks good.

server/src/test/resources/reference.conf
Qi77Qi: these values are mostly just for `ConfigSpec`...no unit test is directly calling any external service..so the actual value doesn't really matter

jdcanas (approved):

PR #148, opened 2022-11-10
IA-3821 do not crash welder when blob not found

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-3821


server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/BackgroundTask.scala
```diff
@@ -93,6 +95,10 @@ class BackgroundTask(
             .through(metadataCacheAlg.updateCachePipe))
             .compile
             .drain
+            .handleErrorWith {
+              case e =>
+                logger.error(loggingCtx, e)(s"Fail to sync cloud storage")
```
Qi77Qi, line 100: this handleErrorWith block is the real change. Basically we shouldn't crash welder when blob can't be found.

n8landolt (approved):
LGTM

There's still references to GCS within the AzureStorageInterp, but I don't think it's necessary to address all of them in this PR

PR #266, opened 2024-02-01
[IA-4806] Updating wblibs version

lucymcnatt:
We recently reverted the cats-effect version in workbench libs to avoid various dependency conflicts that lead to some thread starvation issues. We're need to propagate this change to welder

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -6,7 +6,7 @@ object Dependencies {
   val grpcCoreVersion = "1.51.0"
   val scalaTestVersion = "3.2.14"
 
-  val workbenchLibsHash = "1a6839f"
+  val workbenchLibsHash = "3ae05f8"
   val workbenchGoogle2V = s"0.25-$workbenchLibsHash"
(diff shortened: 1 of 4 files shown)


project/Settings.scala
```diff
@@ -11,7 +11,7 @@ import com.typesafe.sbt.packager.linux.LinuxPlugin.autoImport._
 import sbtbuildinfo.BuildInfoPlugin.autoImport._
 
 object Settings {
-  lazy val artifactory = "https://artifactory.broadinstitute.org/artifactory/"
+  lazy val artifactory = "https://broadinstitute.jfrog.io/artifactory/"
```
lucymcnatt, line 14: needed to update this URL to get welder to recognize the package locations

project/Dependencies.scala
```diff
@@ -1,14 +1,14 @@
 import sbt._
 
 object Dependencies {
```
lucymcnatt, line 3: matched these to the wblibs versions

lucymcnatt:
Not necessary

PR #267, opened 2024-04-08
RW-12258 Update logback

Qi77Qi:
https://precisionmedicineinitiative.atlassian.net/browse/RW-12258

1. We're missing lots of welder logs in cloud logging. This change should make it so that all logs are appended to `CONSOLE` as well
2. Using async-appender similar to what @rtitle did for Leo in https://github.com/DataBiosphere/leonardo/blob/f4e2bf85458be858129b40f1d3ec962bc12015df/http/src/main/resources/logback.xml#L30

### DIFF
--- a/.github/workflows/trivy.yml
+++ b/.github/workflows/trivy.yml
@@ -32,7 +32,8 @@ jobs:
           sbt -no-colors server/docker:publishLocal | tee build.log
 
           # export image name from the log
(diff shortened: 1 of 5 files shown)


LizBaldo (approved):

yonghaoy (approved):

Qi77Qi:
had to do a bit more to get build to pass
1. upgraded workbench libs to what's used in leo today
2. updated jfrog repo reference
3. fixed trivvy GH workflow

PR #270, opened 2024-06-14
ID-1266: Update welder to handle multiple identities assigned to the VM

rtitle:
JIRA: https://broadworkbench.atlassian.net/browse/ID-1266

See: https://github.com/DataBiosphere/leonardo/pull/4629

Notebook VMs might have more than 1 managed identity assigned. We are storing the resource ID of the _pet_ managed in the VM `userData`. When welder needs to get a pet token, first consult the userData for the pet resource ID (if not in userData, default to current behavior).

Tested in a BEE with a mutli-identity VM, and a single-identity VM.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsp/workbench/welder/MiscHttpClientInterp.scala
+++ b/core/src/main/scala/org/broadinstitute/dsp/workbench/welder/MiscHttpClientInterp.scala
@@ -1,8 +1,6 @@
 package org.broadinstitute.dsp.workbench.welder
 
 import cats.effect.IO
-import cats.syntax.all._
-import fs2._
 import org.broadinstitute.dsde.workbench.DoneCheckable
 import org.broadinstitute.dsde.workbench.google2.streamUntilDoneOrTimeout
 import org.broadinstitute.dsp.workbench.welder.MiscHttpClientAlgCodec.{decodePetAccessTokenResp, decodeSasTokenResp}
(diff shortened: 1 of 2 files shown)


Qi77Qi (approved):
the change itself looks good to me...
just to clarify, at any given point, VM won't have multiple identities, but at different time, there might be a different pet manged identity stored in user data?

rtitle:
The VM will have multiple identities assigned: the pet and a dedicated identity for accessing customers’ storage accounts. However azure doesn’t have a notion of a “primary” identity. Hence I’m using the userData to distinguish which identity is the “pet” for welder to talk to WSM.

Btw this won’t be enabled immediately, there is some dependent work in WSM, Leo, etc to set up these identities. This PR is just getting welder ready for when we roll this out.

LizBaldo (commented):
I think I can follow the logic, but would it be possible to add a unit test to make sure that the parsing of the text, and the identity selection works as intended?

rtitle:
Yeah I can add a test, that makes sense.. there is a `MiscHttpClientInterpSpec` but it was a little bare-bones (and I was lazy :)). I can update it

rtitle:
> oh I see...just to check my understanding, we'll have multiple identities for a VM, but welder only cares about PET, and this PR is to make welder knows how to retrieve PET identity instead of other ones.

Yup exactly

rtitle:
@Qi77Qi @LizBaldo I added unit tests (and codecov is happier). Pinging for another review? Thanks 🙏

LizBaldo (approved):

